PR #121, opened 2024-05-15
refactor: add versions & multi-instance docusaurus

FabianKramm:
Fixes DOC-69

### Changes
* Moved vcluster part to `/vcluster`
* Moved platform part to `/platform`
* Removed versioning for overview
* Added versioning for vcluster part (v0.20 & v0.19)
* Added versioning for platform part (v4.0 & v3.4)

Deploy preview https://deploy-preview-121--vcluster-docs-site.netlify.app/docs/


aimeeu:
This is failing due to a lot of broken markdown links (`npm run build`).  I checked out the PR and am working on fixing the links.

aimeeu:
- I fixed the compile errors by changing to relative links, which is [recommended by Docusaurus for versioned docs](https://docusaurus.io/docs/3.2.1/versioning#link-docs-by-file-paths).
- Moved the fragments and partials to vcluster and platform; this is a quick fix and needs to be cleaned up since the platform doesn't use most of the fragments that vcluster does

aimeeu:
> **Get Started** "Get Started" section should be on the 1st level along with "What are virtual clusters" to act as a quick starting point. I understand it will be duplicated inside the vCluster section but currently "What are virtual clusters?" is also duplicated that way and it seems fine.

I moved Get Started to the vCluster section because Get Started contains links to pages int the vCluster section as well as partials. Docusaurus has limitations. 
1. We can't use relative links between docs sets. Overview, vcluster, and platform are all different docs sets. To link to content in a different docs set, we have to use a full URL link not a relative file link. 
2. Partials must live in the docs set that uses them, especially since the partial/fragment is versioned with the docs set.

Also, there are no plans to version the "Overview" landing page. One option would be to make docs site landing page (Overview section) a brief overview of the 2 products rather than repeating the "What are virtual clusters" content.

PR #131, opened 2024-05-30
added example nodeless config

facchettos:
part of doc-60

### DIFF
--- a/platform/integrations/rancher/rancher-integration.mdx
+++ b/platform/integrations/rancher/rancher-integration.mdx
@@ -110,9 +110,12 @@ Disabling the integration at the project level removes all virtual clusters from
 when disabled at this level!
 :::
 
-Depending on your Rancher offering, you may be billed based on the number of nodes that you are using. To avoid duplicated billing,
-you can use the Rancher nodeless plugin, which hides all the nodes from rancher.
-To enable the plugin, use the following config in your vCluster configuration:
+Since virtual clusters are not using physical nodes, you may need to hide the nodes from Rancher in order for the
+virtual nodes to not be counted in your Rancher node count. In order to do so, you will need to enable a plugin as
+part of your virtual cluster deployment.
+
+Add this to your `vcluster.yaml` configuration:
+
 ```yaml
 plugins:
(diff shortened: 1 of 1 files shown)


platform/integrations/rancher/rancher-integration.mdx
```diff
@@ -109,3 +109,16 @@ used to enable it. Disabling the integration at the virtual cluster level simply
 Disabling the integration at the project level removes all virtual clusters from Rancher, so be careful
 when disabled at this level!
 :::
+
+Depending on your Rancher offering, you may be billed based on the number of nodes that you are using. To avoid duplicated billing,
```
deniseschannon, line 113: I don't like the language used in here, but haven't had time to think of a better alternative. 

Let's try this:


Since virtual clusters are not using physical nodes, you may need to hide the nodes from Rancher in order for the virtual nodes to not be counted in your Rancher node count. In order to do so, you will need to enable a plugin as part of your virtual cluster deployment. 

Add this to your `vcluster.yaml` configuration:
```yaml ....
```

Please check our [Rancher plug-in repo](https://github.com/loft-sh/rancher-nodeless-plugin/releases) to always find the latest version available. 

You can take advantage of our [templates](../../templates/) feature and add this configuration change for every virtual cluster deployed through the platform.

platform/integrations/rancher/rancher-integration.mdx
```diff
@@ -109,3 +109,19 @@ used to enable it. Disabling the integration at the virtual cluster level simply
...
+Since virtual clusters are not using physical nodes, you may need to hide the nodes from Rancher in order for the
+virtual nodes to not be counted in your Rancher node count. In order to do so, you will need to enable a plugin as
+part of your virtual cluster deployment.
+
+Add this to your `vcluster.yaml` configuration:
+
+```yaml
+plugins:
+  rancher:
+    name: rancher-plugin
```
FabianKramm, line 122: This can be omitted

FabianKramm (changes requested):

FabianKramm (approved):

PR #142, opened 2024-06-10
Style changes and fixes (DOC-93)

adiron:
- Add right-pointing chevrons to top level product selections
- Change all other chevrons to down/up instead of right/down
- Unify font sizes for different menu items (some where larger?)
- Fix the width gap between the version/product selectors and the rest of the menu
- Add border underneath version/product selectors (to compensate for the lack of background)
- Fix vcluster.yaml logical grouping title styles (more distinct from the other items)
- Fix markdown h<n> line heights

(Corresponds to DOC-93)

### DIFF
--- a/sidebars.js
+++ b/sidebars.js
@@ -19,12 +19,14 @@ const sidebars = {
     {
       "type": "link",
       "label": "vCluster",
-      "href": "/docs/vcluster"
+      "href": "/docs/vcluster",
+      "className": "product-link"
     },
     {
       "type": "link",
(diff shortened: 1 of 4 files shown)


adiron:
Oh no, this is a bug for sure. Something is inverted there in the CSS, and it's somewhat difficult to pull all the threads in the CSS right now. I'll fix this and push an update to this PR.

adiron:
Fixed :)

deniseschannon (approved):

PR #160, opened 2024-06-18
ENG-3943 | Init runme.dev docs test CI

janekbaraniewski:


.github/workflows/test.yml
```diff
@@ -0,0 +1,108 @@
+name: docs-runme
```
FabianKramm, line 1: We should call the file like the run, so either call this `test` or the file `docs-runme.yaml`, I would suggest calling this `test-examples` or `e2e-examples`

FabianKramm (changes requested):

.github/workflows/test.yml
janekbaraniewski: updated 🫡

heiko-braun (approved):
lgtm

heiko-braun (approved):

FabianKramm (approved):

PR #164, opened 2024-06-23
Update to v0.20.0-beta.9

rmweir:
**Issue:**
DOC-102

Install command without flag for chart version does not work with v0.20.0-beta.6.

### DIFF
--- a/vcluster/_partials/deploy/install-beta-cli.mdx
+++ b/vcluster/_partials/deploy/install-beta-cli.mdx
@@ -27,28 +27,28 @@ import TabItem from "@theme/TabItem";
 <TabItem value="mac">
   
   ```bash
-  curl -L -o vcluster "https://github.com/loft-sh/vcluster/releases/download/v0.20.0-beta.6/vcluster-darwin-amd64" && sudo install -c -m 0755 vcluster /usr/local/bin && rm -f vcluster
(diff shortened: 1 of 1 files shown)


vcluster/_partials/deploy/install-beta-cli.mdx
```diff
@@ -27,28 +27,28 @@ import TabItem from "@theme/TabItem";
 <TabItem value="mac">
   
   ```bash
-  curl -L -o vcluster "https://github.com/loft-sh/vcluster/releases/download/v0.20.0-beta.6/vcluster-darwin-amd64" && sudo install -c -m 0755 vcluster /usr/local/bin && rm -f vcluster
+  curl -L -o vcluster "https://github.com/loft-sh/vcluster/releases/download/v0.20.0-beta.9/vcluster-darwin-amd64" && sudo install -c -m 0755 vcluster /usr/local/bin && rm -f vcluster
```
heiko-braun, line 30: This is frequent are that needs change. Would it be better to use a placeholder like `download/<VCLUSTER_VERSION>/vcluster-darwin-amd64`?

Alternatively, is it possible to keep the side wide variables external to the mdx. files, to that we only have to change place?

deniseschannon: We specifically didn't want to put in a random place holder so that people could copy and paste.

deniseschannon (approved):

rmweir:
@heiko-braun I would say after this one we can be less vigilant about updating it until the actual release comes out. This one was important because the command just didn't work as it was until have beta.6.

PR #188, opened 2024-07-10
Update kubeconfig docs to POST instead of GET

zerbitx:


zerbitx:
Converting to draft since I haven't verified the POST command, but need to switch to another task for a while.

deniseschannon (approved):

PR #190, opened 2024-07-10
Eng 3891

zerbitx:
Auto generation pulled in integrations changes as well.


deniseschannon:
@FabianKramm Does the autogeneration happening automatically on `master`? 

This might be a problem cause of how we have to versions for `v0.20` vs. `v0.21`.

deniseschannon (approved):

PR #192, opened 2024-07-11
Specify ns not project

rmweir:
Part of DOC-122

### DIFF
--- a/platform/virtual-clusters/using-virtual-clusters/retrieve-kube-config.mdx
+++ b/platform/virtual-clusters/using-virtual-clusters/retrieve-kube-config.mdx
@@ -15,13 +15,13 @@ The easiest way to retrieve a kubeconfig for your virtual cluster is by using th
 You can update your current kubeconfig on your local machine by running the following command:
 
 ```bash
-vcluster connect my-vcluster --project my-project
+vcluster connect my-vcluster -n my-vclusters-ns
 ```
 
 If you want to print the kubeconfig file, you can append `--print` to the command:
 
 ```bash
-vcluster connect my-vcluster --project my-project --print
+vcluster connect my-vcluster -n my-vclusters-ns --print
 ```
 
 :::info


platform/virtual-clusters/using-virtual-clusters/retrieve-kube-config.mdx
```diff
@@ -15,13 +15,13 @@ The easiest way to retrieve a kubeconfig for your virtual cluster is by using th
 You can update your current kubeconfig on your local machine by running the following command:
 
 ```bash
-vcluster connect my-vcluster --project my-project
```
deniseschannon, line 18: do you not need a flag for project?

rmweir: I explained it in the issue but basically, as far as I can tell, project means nothing to vcluster. totally ignores the flag. Even if I have duplicate names it doesn't disambiguate it.

deniseschannon: So what is `my-project` in this use case? Why even put it in?

rmweir: I'm switch `--project` with `-n` as that does specify it but I accidentally left my-project at the end. Will remove.

deniseschannon: @rmweir I thought I saw another issue that referenced project using the CLI, so now I'm wondering if you were logged into a platform when you performed this. 

vcluster (not logged in) wouldn't care about the project, but if you're logged into the platform, would this be different?

rmweir: @deniseschannon there is a `vcluster platform connect`. I didn't realize because this was just using vcluster. I will try it out and then update docs.

rmweir: Actually there is not `--project` flag for `vcluster platform connect cluster`, it is for `vcluster platform connect namespace`.

rmweir: if you use `vcluster platform connect vcluster <vcluster-name>` and there is more than one vcluster with that name, it will print a sub menu for you to choose the project. I would say that works in this case if you think we should replace everything with `vcluster platform connect vcluster <vcluster-name>`.

rmweir:
Closed in favor of https://github.com/loft-sh/vcluster-docs/pull/201.

PR #194, opened 2024-07-16
Adds external.platform.(apiKey/autoSleep/autoDelete) docs

zerbitx:
Resolves DOC-192 and DOC-196

vCluster preview: https://deploy-preview-194--vcluster-docs-site.netlify.app/docs/vcluster/configure/vcluster-yaml/
vCluster Platform preview: https://deploy-preview-194--vcluster-docs-site.netlify.app/docs/platform/virtual-clusters/configure/vcluster-yaml/


zerbitx:
Setting this as draft until the createRBAC apiKey changes are merged and I can update.

zerbitx:
I talked with Russ and Denise and denise about this one and where the platform config options should reside given that the vcluster.yaml docs are on the OSS side of the docs.  The plan is to have the generator produce the partials in the regular location and only have it the platform specific fields show up in a new `vcluster.yaml` section on the platform side.

rohantmp (commented):
It's not obvious to me if fields like autoSleep and external are top level vcluster.yaml fields from this page: https://deploy-preview-194--vcluster-docs-site.netlify.app/docs/platform/api/resources/vcluster-yaml/

I know external is, so I assume autoSleep is too from the structure of the docs, but maybe we can add a comment on the vcluster.yaml page in platform to make this clearer?

zerbitx:
@rohantmp very good point.  I've updated to build the partials for the hierarchy and put that at the top level

deniseschannon:
Shouldn't there be updates on vcluster.yaml and the support in there? I don't see it in this PR.

zerbitx:
@deniseschannon Added those to the vCluster side with pro labels.

deniseschannon:
Based on the files, it looks like we're kind of autogenerating it, but I'm not sure.

The generated partials include information from the config file that describe the key. 

Example:
Screenshot 2024-07-19 at 5 31 48 AM

Where as the current keys in the external field are all blank:

Screenshot 2024-07-19 at 5 31 41 AM

Can we add that info in for the external keys as well?

deniseschannon:
The layout for the new pages look different than the older pages, where we generally have a small blurb about what the key is about. A simple explanation of how to use it. Then we have an example section and finally the config reference. 

https://deploy-preview-194--vcluster-docs-site.netlify.app/docs/vcluster/configure/vcluster-yaml/experimental/deny-proxy-requests

deniseschannon:
Somehow we're missing pages for `external` and the `platform` levels. You can see in this page, how the toggle navbar is actually also a link. 

It would be good for us to explain what anything under external means and what does platform. If you need help with that explanation, let me know. 

https://deploy-preview-194--vcluster-docs-site.netlify.app/docs/vcluster/configure/vcluster-yaml/sync/

deniseschannon:
Platform Comments 

I'm fine with adding this in resources, but also think we need somewhere in the Virtual Cluster section which describes how externally deployed vclusters could connect to the platform. Unfortunately, I haven't gotten around to adding in content about externally deployed vs. platform deployed. But if we add a new page for "Connecting to the Platform" that might be a good starting point.

_partials/what-is-vcluster-yaml.mdx
zerbitx, line 1: No _need_ to re-review this as it was just copied from the oss docs into a partial without modification, save for omitting the head matter.

deniseschannon:
This is difficult for reviews, so I'll merge as is and update it afterwards.

deniseschannon (approved):

PR #209, opened 2024-07-25
Add ns to OIDC client secret example

rmweir:


platform/manage/oidc-provider.mdx
```diff
@@ -33,6 +33,7 @@ apiVersion: v1
 kind: Secret
 metadata:
   name: test-oidc-client
+  namespace: vcluster-pltform # this should be which ever namespace vCluster Platform is installed in
```
deniseschannon, line 36: typo in `vcluster-pltfrm`

The comment should be `# The namespace where there vCluster Platform is installed in`

PR #211, opened 2024-07-29
HA with embedded SQLite is not supported

rmweir:
Part of ENG-4076

Preview link: https://deploy-preview-211--vcluster-docs-site.netlify.app/docs/vcluster/deploy/topologies/high-availability/#create-the-ha-virtual-cluster

### DIFF
--- a/vcluster/deploy/topologies/high-availability.mdx
+++ b/vcluster/deploy/topologies/high-availability.mdx
@@ -27,6 +27,9 @@ controlPlane:
     highAvailability:
       replicas: 3
 ```
+:::info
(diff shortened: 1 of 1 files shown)


vcluster/deploy/topologies/high-availability.mdx
```diff
@@ -27,6 +27,10 @@ controlPlane:
     highAvailability:
       replicas: 3
 ```
+:::info
```
deniseschannon, line 30: Instead of giving all the other options, maybe we simplify it to:
```
High Availability mode is not compatible with the embedded SQLite as the backing store, which is the default backing store. You must explicitly set an alternative [backing store](../../configure/vcluster-yaml/control-plane/components/backing-store).
```

deniseschannon (changes requested):

deniseschannon (approved):

PR #216, opened 2024-08-05
[0.21] fix: observability deprecated in beta.12

Smidra:
Documentation should reflect changes made in https://github.com/loft-sh/vcluster/pull/1903
`observability --> integrations`


deniseschannon:
Thanks for the PR, that's only in `main`, which is not where beta.12 is cut from. Instead that's our next v0.21 release, so I'll keep this PR open for it.

FabianKramm (approved):

PR #217, opened 2024-08-05
add example of using dataVolume chart value to re-use existing PVC

hidalgopl:
resolves ENG-4166

### DIFF
--- a/vcluster/configure/vcluster-yaml/control-plane/deployment/statefulset.mdx
+++ b/vcluster/configure/vcluster-yaml/control-plane/deployment/statefulset.mdx
@@ -30,6 +30,20 @@ The vCluster control plane is deployed as a StatefulSet, since vCluster requires
 - `topologySpreadConstraints`: See [Pod Topology Spread Constraints](https://kubernetes.io/docs/concepts/scheduling-eviction/topology-spread-constraints/).
   - Similar to podAffinity & podAntiAffinity in some ways. See [Comparison with podAffinity and podAntiAffinity](https://kubernetes.io/docs/concepts/scheduling-eviction/topology-spread-constraints/#comparison-with-podaffinity-podantiaffinity).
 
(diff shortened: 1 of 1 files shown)


johannesfrey (approved):

vcluster/configure/vcluster-yaml/control-plane/deployment/statefulset.mdx
```diff
@@ -30,6 +30,20 @@ The vCluster control plane is deployed as a StatefulSet, since vCluster requires
 - `topologySpreadConstraints`: See [Pod Topology Spread Constraints](https://kubernetes.io/docs/concepts/scheduling-eviction/topology-spread-constraints/).
   - Similar to podAffinity & podAntiAffinity in some ways. See [Comparison with podAffinity and podAntiAffinity](https://kubernetes.io/docs/concepts/scheduling-eviction/topology-spread-constraints/#comparison-with-podaffinity-podantiaffinity).
 
+## Re-using existing PersistenceVolumeClaim
+It is possible to pre-provision a PersistenceVolumeClaim, and then configure vCluster to use it.
+To do that, set existing claim name in the chart values `controlPlane.statefulSet.persistence.dataVolume` under `persistenceVolumeClaim.claimName`.
```
deniseschannon, line 35: I believe this changes it from a StatefulSet to Deployment for the control plane, which is something we may want to add in as part of the docs. 

Can you double check that and add a note about it if so?

hidalgopl: It doesn't change it automatically, I think additional conditions has to be met to make it Deployment, see https://github.com/loft-sh/vcluster/blob/main/chart/templates/_persistence.tpl#L5 & https://github.com/loft-sh/vcluster/blob/main/chart/templates/_persistence.tpl#L33-L41

deniseschannon (approved):

PR #238, opened 2024-08-13
Adding changed to yaml for embedded etcd

sowmyav27:
Fixes Linear issue: `DOC-224`

Existing docs:
- https://www.vcluster.com/docs/vcluster/configure/vcluster-yaml/control-plane/components/backing-store/etcd/embedded

Pages fixed:
- https://deploy-preview-238--vcluster-docs-site.netlify.app/docs/vcluster/configure/vcluster-yaml/control-plane/components/backing-store/etcd/embedded

### DIFF
--- a/vcluster/configure/vcluster-yaml/control-plane/components/backing-store/etcd/embedded.mdx
+++ b/vcluster/configure/vcluster-yaml/control-plane/components/backing-store/etcd/embedded.mdx
@@ -13,14 +13,17 @@ import ProAdmonition from '@site/vcluster/_partials/admonitions/pro-admonition.m
 
 etcd is deployed with vCluster to reduce the overall footprint while still providing the advantages of etcd.
 
-You can migrate from an external etcd instance to the embedded one by configuring `migrateFromDeployedEtcd: true`.
+You can migrate from an external etcd instance to the embedded one by configuring `migrateFromDeployedEtcd: true` as shown below. However, by default, `enabled` is set to `false`, and `migrateFromDeployedEtcd` is also set to `false`.
+
+
 
 ```yaml
 controlPlane:
   backingStore:
     etcd:
       embedded:
-        ...
+        enabled: true 
+        migrateFromDeployedEtcd: true 
 ```
 
 :::warning


vcluster/configure/vcluster-yaml/control-plane/components/backing-store/etcd/embedded.mdx
```diff
@@ -20,7 +20,8 @@ controlPlane:
   backingStore:
     etcd:
       embedded:
-        ...
+        enabled: false # default value
+        migrateFromDeployedEtcd: false #default value
```
rmweir, line 24: I think if this must be placed under "You can migrate from an external etcd instance to the embedded one by configuring `migrateFromDeployedEtcd: true`." Then this should be true. Another option would be to keep this false and then change the line above to be a note either below or above so its clearly a suggestion in addition to what is being presented generally.

zerbitx (approved):

vcluster/configure/vcluster-yaml/control-plane/components/backing-store/etcd/embedded.mdx
```diff
@@ -13,14 +13,17 @@ import ProAdmonition from '@site/vcluster/_partials/admonitions/pro-admonition.m
 
 etcd is deployed with vCluster to reduce the overall footprint while still providing the advantages of etcd.
 
-You can migrate from an external etcd instance to the embedded one by configuring `migrateFromDeployedEtcd: true`.
+You can migrate from an external etcd instance to the embedded one by configuring `migrateFromDeployedEtcd: true` as shown below. However, by default, `enabled` is set to `false`, and `migrateFromDeployedEtcd` is also set to `false`.
```
deniseschannon, line 16: I'm not sure if you need any additional information. Did you test this out? 

But otherise, I'd change it to:

```
You can migrate from an external etcd instance to the embedded one by configuring `migrateFromDeployedEtcd: true`. You will also need to enable the embedded etcd in order to enable the migration. 
```

deniseschannon (changes requested):

PR #240, opened 2024-08-14
vcluster cli throws an error if exportKubeConfig.server does not have a port number when set

brinker-tbaker:
when editing the exportKubeConfig.server property you have to specify the port number as well, otherwise an error is thrown from the vcluster cli as there are not 2 instances of a : in the path. This change updates the documentation to include the port number.

see [pkg/cli/connect_helm.go](https://github.com/loft-sh/vcluster/blob/450e5288b68b728222acf5d0046b529a79ac6c8d/pkg/cli/connect_helm.go#L372)
```
splitted := strings.Split(cluster.Server, ":")
if len(splitted) != 3 {
   return nil, fmt.Errorf("unexpected server in kubeconfig: %s", cluster.Server)
}
```

### DIFF
--- a/vcluster/manage/accessing-vcluster.mdx
+++ b/vcluster/manage/accessing-vcluster.mdx
@@ -161,7 +161,7 @@ controlPlane:
     scheduling:
       podManagementPolicy: OrderedReady
 exportKubeConfig:
-  server: https://my-domain.org
+  server: https://my-domain.org:443
 ```
 
 Then you can create or upgrade the vCluster with:


deniseschannon:
@brinker-tbaker Can you confirm what version of the CLI that you're using?

brinker-tbaker:
@deniseschannon 
```
vcluster --version
vcluster version 0.20.0-beta.15
```
If the 0.20.0 release should NOT require a port number, i can submit a PR on the main repo for that instead if needed.

brinker-tbaker:
upon further review, it looks like this is an edge case if you specify `server: https://path.to.my.server` but works correctly if you just put `server: path.to.my.server`. This can be closed out and I'll open a PR for the code so that all 3 formats work  fqdn, fqdn:port, and schema:fqdn:port.

zerbitx (approved):

FabianKramm (approved):

PR #251, opened 2024-09-02
add section about additionalCA

hidalgopl:
resolves ENG-4224

@matskiv @deniseschannon let me know if you think there's a better place in the docs to put this

@hidalgopl Next time, you can take the deploy preview and add it into comment field so it's easier to review:

https://deploy-preview-251--vcluster-docs-site.netlify.app/docs/platform/install/configure/config#setting-additional-certificate-authority-for-platform

### DIFF
--- a/platform/install/configure/config.mdx
+++ b/platform/install/configure/config.mdx
@@ -93,6 +93,30 @@ config:
       X-XSS-Protection: 1; mode=block
 ```
 
+### Setting additional certificate authority for Platform
+If you decide to configure vCluster Platform helm release with  `additionalCA`, it will be automatically used by all vClusters managed by the vCluster Platform to establish a connection between vClusters and the Platform. You don't need to tweak individual vCluster configuration to enable this.
+This will enable secure communication between your vClusters and the vCluster Platform.
+```yaml
+additionalCA: "(base64 encoded CA)"
+```
+
+Also, if you need to import externally managed vCluster to the vCluster Platform, you can also enable a secure connection by grabbing the value from the  existing vCluster Platform helm release, you can do this by running:
+
+```shell
+helm get values --namespace [namespace where vCluster Platform is deployed] [loft helm release name] --all | grep additionalCA
+```
+
+Then, using you can import vCluster using vCluster CLI command:
+```shell
+  vcluster platform add vcluster [your vCluster name] \
+--namespace=[namespace where vCluster runs] \
+--project=[project name] \
+--ca-data [base64 string grabbed from the command above]
+```
+
+Once the command succeeds, imported vCluster should restart and report Ready phase.
+
+
 ### The vCluster Platform Agent
 
 Every connected cluster in a vCluster Platform environment, including the "main" cluster that vCluster Platform is
--- a/platform/install/configure/config.mdx
+++ b/platform/install/configure/config.mdx
@@ -94,27 +94,27 @@ config:
 ```
 
 ### Setting additional certificate authority for Platform
-If you decide to configure vCluster Platform helm release with  `additionalCA`, it will be automatically used by all vClusters managed by the vCluster Platform to establish a connection between vClusters and the Platform. You don't need to tweak individual vCluster configuration to enable this.
-This will enable secure communication between your vClusters and the vCluster Platform.
+If you want to have **all** your virtual clusters managed by the vCluster Platform to use a specific certificate authority to establish the connection between the virtual cluster and the platform, you can set it as a configuration option on the vCluster Platform instead of configuring it on each individual virtual cluster.
+This will enable secure communication between your Virtual Clusters and the vCluster Platform.
 ```yaml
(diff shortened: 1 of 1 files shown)


platform/install/configure/config.mdx
```diff
@@ -93,6 +93,30 @@ config:
       X-XSS-Protection: 1; mode=block
 ```
 
+### Setting additional certificate authority for Platform
+If you decide to configure vCluster Platform helm release with  `additionalCA`, it will be automatically used by all vClusters managed by the vCluster Platform to establish a connection between vClusters and the Platform. You don't need to tweak individual vCluster configuration to enable this.
```
deniseschannon, line 97: Update to: 
```
If you want to have **all** your virtual clusters managed by the vCluster Platform to use a specific certificate authority to establish the connection between the virtual cluster and the platform, you can set it as a configuration option on the vCluster Platform instead of configuring it on each indiviudal virtual cluster.
```

Note about the docs: You can't pluralize vCluster so it must be virtual clusters.

platform/install/configure/config.mdx
```diff
@@ -93,6 +93,30 @@ config:
...
+
+Then, using you can import vCluster using vCluster CLI command:
+```shell
+  vcluster platform add vcluster [your vCluster name] \
+--namespace=[namespace where vCluster runs] \
+--project=[project name] \
+--ca-data [base64 string grabbed from the command above]
+```
+
+Once the command succeeds, imported vCluster should restart and report Ready phase.
```
deniseschannon, line 117: Put Ready into a code block `Ready`. Personally, I hate the word "should" in docs and it should be more assertive. 

```
Once the command succeeds, the imported vCluster pod restarts and reports `Ready`.
```

platform/install/configure/config.mdx
```diff
@@ -93,6 +93,30 @@ config:
       X-XSS-Protection: 1; mode=block
 ```
 
+### Setting additional certificate authority for Platform
+If you decide to configure vCluster Platform helm release with  `additionalCA`, it will be automatically used by all vClusters managed by the vCluster Platform to establish a connection between vClusters and the Platform. You don't need to tweak individual vCluster configuration to enable this.
+This will enable secure communication between your vClusters and the vCluster Platform.
```
deniseschannon, line 98: Update vClusters to virtual clusters.

platform/install/configure/config.mdx
```diff
@@ -93,6 +93,30 @@ config:
...
+```yaml
+additionalCA: "(base64 encoded CA)"
+```
+
+Also, if you need to import externally managed vCluster to the vCluster Platform, you can also enable a secure connection by grabbing the value from the  existing vCluster Platform helm release, you can do this by running:
```
deniseschannon, line 103: We don't call installing vCluster Platform as the `vCluster Platform helm release`, but this something to figure out how we want to refer to it. 

For now, I would call it vCluster Platform until we standardize it.

platform/install/configure/config.mdx
```diff
@@ -93,6 +93,30 @@ config:
...
+
+```shell
+helm get values --namespace [namespace where vCluster Platform is deployed] [loft helm release name] --all | grep additionalCA
+```
+
+Then, using you can import vCluster using vCluster CLI command:
```
deniseschannon, line 109: Update to
`Using the `additionalCA` value from the above command, import the vCluster using the following vCluster CLI command:

platform/install/configure/config.mdx
```diff
@@ -93,6 +93,30 @@ config:
...
+```shell
+  vcluster platform add vcluster [your vCluster name] \
```
deniseschannon, line 111: Typically when we do commands that require replace statements we have a style that is like

`<VCLUSTER_NAME>`
`<VCLUSTER_NAMESPACE>`
`<PROJECT_NAME>`

And then after the command it gives a hint that you should replace those values with a description of what they mean. 

Example of what I mean is shown here: https://www.vcluster.com/docs/vcluster/manage/upgrade-version?x0=1#upgrade-vcluster-version

deniseschannon (changes requested):

platform/install/configure/config.mdx
hidalgopl: :+1: changed

platform/install/configure/config.mdx
hidalgopl: updated!

platform/install/configure/config.mdx
hidalgopl: updated :+1:

platform/install/configure/config.mdx
hidalgopl: ok, removed `helm release`

platform/install/configure/config.mdx
hidalgopl: :+1: updated

deniseschannon (approved):

PR #255, opened 2024-09-06
Update replicate-services.mdx add networking in yaml snippets

thedocks:
add missing networking parent in yaml snippets


deniseschannon (approved):
Thanks @thedocks!

PR #258, opened 2024-09-11
added guide for eso

facchettos:
resolves doc-240

### DIFF
--- a/vcluster/configure/vcluster-yaml/integrations/metricsServer.mdx
+++ b/vcluster/configure/vcluster-yaml/integrations/metricsServer.mdx
@@ -4,6 +4,6 @@ sidebar_label: metricsServer
 sidebar_position: 1
 ---
 
-import MetricsServer from '@site/vcluster/_fragments/metrics-server.mdx'
+import MetricsServer from '@site/vcluster/_fragments/eso.mdx'
 
 <MetricsServer />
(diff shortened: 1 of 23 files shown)


vcluster/integrations/external-secrets/guide.mdx
```diff
@@ -0,0 +1,89 @@
...
+  - extract:
+      key: /foo/baz
+      version: v1beta1
+```
+
+Once that external secret is created in the virtual cluster, the integration will take care or creating a corresponding secret inside the host cluster,
+the external secret operator running in the host will take care of creating the corresponding kubernetes secret, and the integration will import this
+secret in the virtual cluster. Running `kubectl get secrets` in the virtual cluster should now include the `secret-to-be-created` in its output.
+
+
```
FabianKramm, line 89: Would be good to add the config reference here like we do for other pages

FabianKramm (commented):
Can we also add that to the vcluster.yaml -> integrations -> externalSecrets similar to kubeVirt / metricsServer?

vcluster/configure/vcluster-yaml/integrations/metricsServer.mdx
```diff
@@ -4,6 +4,6 @@ sidebar_label: metricsServer
 sidebar_position: 1
 ---
 
-import MetricsServer from '@site/vcluster/_fragments/metrics-server.mdx'
```
FabianKramm, line 7: This looks wrong

FabianKramm (changes requested):

FabianKramm (approved):

PR #284, opened 2024-10-04
add docs on how to get a value for additionalCA

hidalgopl:
This links existing TLS configuration section with additionalCA, so it's clear how to get the value for additionalCA option.


Piotr1215 (approved):
LGTM!

PR #312, opened 2024-10-16
feat: expand vale check to vcluster and vcluster_versioned folders (#238)

Piotr1215:
Closes DOC-238


ThomasK33 (dismissed):

.github/workflows/reviewdog.yaml
ThomasK33, line 1: It's more of a nit, but we might want to drop the "platform" part from the job name in line 42 now that it's not platform-exclusive anymore.

ThomasK33 (approved):

PR #315, opened 2024-10-17
fix: reviewdog token access

Piotr1215:
Closes DOC-303


.github/workflows/reviewdog.yaml
```diff
@@ -67,7 +67,7 @@ jobs:
         with:
           token: ${{ secrets.GH_ACCESS_TOKEN }}
```
ThomasK33, line 68: We might also try changing the token here to 

```suggestion
          token: ${{ secrets.GITHUB_TOKEN }}
```

ThomasK33 (approved):

PR #317, opened 2024-10-17
fix: file detection

Piotr1215:
Closes DOC-303

### DIFF
--- a/.github/workflows/reviewdog.yaml
+++ b/.github/workflows/reviewdog.yaml
@@ -35,7 +35,7 @@ jobs:
       - name: Run Misspell with Reviewdog
         uses: reviewdog/action-misspell@v1
         with:
-          github_token: ${{ secrets.GH_ACCESS_TOKEN }}
+          github_token: ${{ secrets.GITHUB_TOKEN }}
           reporter: github-pr-review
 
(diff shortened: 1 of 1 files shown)


.github/workflows/reviewdog.yaml
```diff
@@ -35,7 +35,7 @@ jobs:
       - name: Run Misspell with Reviewdog
         uses: reviewdog/action-misspell@v1
         with:
-          github_token: ${{ secrets.GH_ACCESS_TOKEN }}
+          github_token: ${{ secrets.GITHUB_TOKEN }}
           reporter: github-pr-review
```
ThomasK33, line 39: This will probably have to become a `github-pr-check`

Piotr1215: It works also with `github-pr-review` and I think comments are a bit more readable.

ThomasK33: Oh, in that case we might want to change back the vale linter to review comments

Piotr1215: Good call, thank you! I forgot to add it.

ThomasK33 (approved):

PR #336, opened 2024-10-28
Update wrong release url

mjudeikis:
Not sure why its different the usual ones where downloads are after the tag :/

Found this while trying to install.


Piotr1215 (approved):
Thank you for catching this @mjudeikis!

PR #346, opened 2024-11-05
feat: clarify supported kubernetes versions for airgapped installation

Piotr1215:
Closes DOC-334

Docs Preview: https://deploy-preview-346--vcluster-docs-site.netlify.app/docs/platform/install/advanced/air-gapped#supported-kubernetes-versions

### DIFF
--- a/platform/install/advanced/air-gapped.mdx
+++ b/platform/install/advanced/air-gapped.mdx
@@ -41,8 +41,11 @@ accessible to your cluster.
 
 ### Supported Kubernetes versions
 
-<!-- TODO:(piotr1215) figure out which versions are supported -->
-- A running air-gapped Kubernetes cluster `v1.26.x`, `v1.27.x`, or `v1.28.x`
+- A running air-gapped Kubernetes cluster with one of the supported versions: `v1.30.x`, `v1.31.x`, or `v1.32.x`
+
+:::note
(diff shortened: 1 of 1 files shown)


deniseschannon (approved):

platform/install/advanced/air-gapped.mdx
```diff
@@ -41,8 +41,11 @@ accessible to your cluster.
 
 ### Supported Kubernetes versions
 
-<!-- TODO:(piotr1215) figure out which versions are supported -->
-- A running air-gapped Kubernetes cluster `v1.26.x`, `v1.27.x`, or `v1.28.x`
+- A running air-gapped Kubernetes cluster with one of the supported versions: `v1.30.x`, `v1.31.x`, or `v1.32.x`
```
deniseschannon, line 44: Now, technically, we've never tested out v1.32, but to prevent us always trying to update these docs. Should we put a generic statement of "supported versions of k8s" and refer to the k8s support matrix?  

This would allow less maintenance for us. WDYT?

Piotr1215: I was hoping this would be possible, but the version support depends on the `client-go` library in the platform. This is currently on `k8s.io/client-go v0.31.1` and can change in the future versions.

PR #348, opened 2024-11-08
feat: version badge component

Piotr1215:
Closes DOC-339


deniseschannon:
I wonder if we need to also indicate what version of vcluster this was introduced in.

Piotr1215:
Good idea, I've added an optional `vclusterVerson` property which would render like so:

image

johannesfrey (approved):

FabianKramm (approved):

PR #359, opened 2024-11-18
docs: adding info about etcd migration path for pro users

Piotr1215:
Closes DOC-326

[Backing Store Preview](https://deploy-preview-359--vcluster-docs-site.netlify.app/docs/vcluster/next/configure/vcluster-yaml/control-plane/components/backing-store/)

### DIFF
--- a/vcluster/configure/vcluster-yaml/control-plane/components/backing-store/etcd/deploy.mdx
+++ b/vcluster/configure/vcluster-yaml/control-plane/components/backing-store/etcd/deploy.mdx
@@ -7,6 +7,9 @@ description: Configure an external etcd instance as the virtual cluster's backin
 
 import ConfigReference from '../../../../../../_partials/config/controlPlane/backingStore/etcd/deploy.mdx'
 
+import ProAdmonition from '../../../../../../_partials/admonitions/pro-admonition.mdx';
+
+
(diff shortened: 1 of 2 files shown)


vcluster/configure/vcluster-yaml/control-plane/components/backing-store/README.mdx
```diff
@@ -17,6 +17,11 @@ vCluster is deployed using a StatefulSet (set of pod replicas + persistent volum
 After deploying your vCluster, changing the backing store of vCluster is not supported.
 :::
 
+For vCluster Pro users, it is possible to migrate the backing store of a virtual
+cluster, with only [etcd](https://etcd.io/) supported as a target backing store.
+
```
deniseschannon, line 22: This implies that we could migrate from any backing store to etcd, but we only support going from `controlPlane.backingstore.etcd.deploy` to `controlPlane.backingstore.etcd.embedded`

I wouldn't add it in the generic Backing store page, but add this info into the deploy etcd page that this is an option. 

https://www.vcluster.com/docs/vcluster/configure/vcluster-yaml/control-plane/components/backing-store/etcd/deploy

Also, I wouldn't worry about backporting these into the older versions before the latest stable version. 

I'd rather focus on continuing to improve the current stable version and the next.

deniseschannon (changes requested):

vcluster/configure/vcluster-yaml/control-plane/components/backing-store/README.mdx
Piotr1215: I've removed the other changes and added the info to the `etcd-deploy` page.

deniseschannon (approved):

PR #398, opened 2024-12-18
do not hardcode partials paths in vCluster partials generation script

hidalgopl:
instead of hardcoding needed paths, generate paths dynamically by traversing the schema tree

# Content Description

## Internal Reference

<!--Add the Github or Linear ticket reference>
Closes

### DIFF
--- a/hack/vcluster/partials/main.go
+++ b/hack/vcluster/partials/main.go
@@ -5,94 +5,16 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
+
+	orderedmap "github.com/wk8/go-ordered-map/v2"
 
 	"github.com/ghodss/yaml"
 	"github.com/invopop/jsonschema"
 	"github.com/loft-sh/vcluster-docs/hack/platform/util"
 )
 
-// we only generate paths we actually need
-var paths = []string{
-	"sync/fromHost/customResources",
-	"sync/toHost/customResources",
-	"policies/podSecurityStandard",
-	"telemetry",
-	"sync/toHost/volumeSnapshots",
-	"sync/toHost/storageClasses",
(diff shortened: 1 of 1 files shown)


hack/vcluster/partials/main.go
```diff
@@ -3,95 +3,17 @@ package main
...
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/ghodss/yaml"
 	"github.com/invopop/jsonschema"
 	"github.com/loft-sh/vcluster-docs/hack/platform/util"
 )
 
-// we only generate paths we actually need
```
Piotr1215, line 14: I wonder what was the reason for us to keep this slice manual? Is there a scenario where we would _not_ want to generate full partials? cc @FabianKramm

hack/vcluster/partials/main.go
```diff
@@ -123,7 +45,44 @@ func main() {
...
+}
+
+func walkTree(node, parent *jsonschema.Schema, name, parentName string) bool {
+	if node == nil || getChildren(node, parent) == nil {
+		return true
+	} else {
+		parentName = fmt.Sprintf("%s/%s", parentName, name)
+		paths = append(paths, parentName)
+	}
+	children := getChildren(node, parent)
```
Piotr1215, line 65: Small nit, we should avoid else after return and exit early for readability, maybe something like:

```go
    if node == nil {
        return true
    }
    
    children := getChildren(node, parent)
    if children == nil {
        return true
    }
```

Piotr1215 (commented):
Thank you, just one question to @FabianKramm and small nit.

Piotr1215:
Closing this as won't do. After talking with Fabian, we would like to keep the control over generating the menu items manually.

PR #425, opened 2025-01-17
feat: Add vale rule to block use of site in local partials

nfrankel:
# Content Description

Add vale rule to block use of site in local partials. Also fix existing partials to conform to the rule.

Didn't fix usage of fragments because there were too many existing violations.

>We want to avoid using `@site` directive to reference docs specific `_partials` (references inside vcluster, platform or versioned folders). 
>
>To make this check automated, we would like to a create a vale rule to capture and enforce the logic. 

## Preview link

https://deploy-preview-425--vcluster-docs-site.netlify.app/docs/

## Internal Reference
Closes DOC-287

### DIFF
--- a/.github/styles/Loft/centralize-partials-fragments.yml
+++ b/.github/styles/Loft/centralize-partials-fragments.yml
@@ -1,0 +1,6 @@
+extends: existence
+message: "Avoid using @site directive for referencing docs-specific partials and fragments. Use relative paths instead."
+level: error
+scope: raw
+raw:
+  - '@site/(?:vcluster|platform|.*versioned.*)/[_partials|_fragments]'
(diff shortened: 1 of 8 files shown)


Piotr1215 (commented):
We should also add the same rule for the `_fragments` folders, as they are used with partials.

`@site` calls should be allowed for the `docs` folder in the repo root (which should already be ok with the rule):

```bash
.
├── get-started.mdx
├── index.mdx
└── _partials
    ├── kubeconfig_update.mdx
    └── managed_k8s_install_vcluster.mdx

1 directory, 4 files
```

It's good to test the rule in the [vale studio](https://studio.vale.sh/) and provide a sharable test result link in the PR.

Piotr1215 (approved):
Thank you for the PR @nfrankel.

Adding vale studio tests: https://studio.vale.sh/s/f028e90494c82779de89c8acf28de041

image

PR #434, opened 2025-01-23
Updates lifecycle page and k8s compatability

cbron:
# Content Description
Converts lifecycle's latest release to link instead of number

## Internal Reference

None


Piotr1215 (approved):
@cbron if you are ok with it, let's merge this PR so we can implement DOC-403, otherwise there would be merge conflicts.

PR #442, opened 2025-01-27
Adjust title headings

nebelsprachen:
# Content Description
- Some of the titles use a mix of title case and sentence case. Since we are treating "Platform" as a proper noun, I left it capitalized. However, other terms like "integration" or "configuration" are not proper nouns, so they should likely be in lowercase to align with sentence case style.

- Slight rewrite to text and added commas for clarity

> **Note**: The term "Platform" is not formatted consistently throughout the document. I opted to use it as a proper noun to align with the document's headings ( but it could go either way). Additionally, according to [the Google style guide](https://developers.google.com/style/ui-elements), UI elements should be formatted in bold typeface. There are a few instances where this formatting should be updated to match the style guide.

## Preview Link 
- [Update headings and text](https://deploy-preview-442--vcluster-docs-site.netlify.app/docs/platform/install/gitops)
- [Update headings for sentence case](https://deploy-preview-442--vcluster-docs-site.netlify.app/docs/platform/configure/platform-config/) 
- [Add a comma](https://deploy-preview-442--vcluster-docs-site.netlify.app/docs/platform/integrations/rancher/rancher-integration)
- Fix GitHub format typo

## Internal Reference
n/a

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -6,6 +6,6 @@
 If you want the preview link to be available in the Linear issue, you must include the word `preview` in the markdown link name [Document Preview](https://netlify.preview/docs/xxxx). -->
 
 ## Internal Reference
-<!--Add the Github or Linear ticket reference-->
+<!--Add the GitHub or Linear ticket reference-->
 Closes DOC-
 
--- a/platform/configure/config.mdx
+++ b/platform/configure/config.mdx
@@ -1,6 +1,6 @@
 ---
 title: Platform configuration
-sidebar_label: Platform config
+sidebar_label: Platform Config
 sidebar_position: 1
 description: Learn how to configure the vCluster Platform, modify its behavior, and enable or disable features such as SSO login, custom branding, and auditing.
 ---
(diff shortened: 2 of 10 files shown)


Piotr1215 (changes requested):
@nebelsprachen I had to approve/trigger the `vale` workflow (it needs to be done for all external contributors). As you can see, there are some `vale` rules violations. Could you please fix those? Please refer to the CONTRIBUTING.md to learn more about styling guidelines and working with vale.

> [!TIP]
>  you can find the exact rules violations in the `Files changed` tab

nebelsprachen:
Thank you, @Piotr1215. I addressed some of the linter comments, but others seem more like suggestions rather than strict rules and may not be entirely correct. For example, one rule states that "admin" should not be capitalized, but in the given example, the text references a UI element, where capitalization seems be appropriate.

Another rule suggests using "the platform" instead of "the vCluster Platform," but this change could make the text ambiguous, especially for new users who may not be familiar with the product. 

Also, mentioned in the comment above "platform" is formatted inconsistently throughout the docs. I am not sure if it is a proper noun, as in "the vCluster Platform" or if we are just referring to it as "the  platform".

Piotr1215:
## Vale rules override

In the case of `admin` not being capitalized, the CONTRIBUTING.md has a chapter dedicated to [controlling vale issues](https://github.com/loft-sh/vcluster-docs/blob/main/CONTRIBUTING.md#controlling-vale-rules).

## Product naming
The product naming should be `the platform`, however front matter is an exception here where we want to improve the discoverability by search engines. Front matter is exempt from value rules.

> [!NOTE]
> add the above to the style guidelines

## Inconsistent formatting

Not all documents were changed since I've introduced `vale`, so there are some older docs that will still have rules violations, and they will need to be fixed either when the doc is being changed or via spot-checks.

Piotr1215 (approved):
Thank you @nebelsprachen LGTM!

PR #443, opened 2025-01-27
chore: add info about using vale

Piotr1215:
# Content Description

Adding info about dealing with `vale` errors to the PR template.

## Internal Reference

Closes DOC-407

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,9 +1,17 @@
+<!-- 
+When changing something in a file, our linting system `vale`, will treat the whole file as changed and will lint it. 
+In this case, follow the instructions from vale and fix the linting issues. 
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -1,10 +1,18 @@
+<!--
```
deniseschannon, line 1: I think the downside of leaving these as comments is that it's not seen by default when you create a PR. Not sure how to fix that, but otherwise it looks good.

deniseschannon (approved):

.github/pull_request_template.md
Piotr1215: I think the comments are visible in the edit mode, but maybe there are settings how edit mode looks? That's what I see when I create new PR:
image

PR #447, opened 2025-01-31
Improved PRO label drawing, and fixed a display bug

adiron:
The PRO label is now applied to the `::after` pseudo-selector of the actual `.pro` element. This is to prevent a collision with Docusaurus' own `::after`, which draws the submenus generally.

This affects the way that the PRO label is positioned (added comments for future generations).

Additionally, I have implemented some changes for a more consistent look for the PRO tags across different sections (though, some small and mysterious quirks of style do remain)

# Content Description

## Internal Reference

Closes DOC-397

## Screenshots

Support for collapsible pro items (**none currently exist, this is a test**):
Screenshot 2025-01-31 at 12 39 09

Other pro tags:
image
image

### DIFF
--- a/src/css/sidebar.scss
+++ b/src/css/sidebar.scss
@@ -10,36 +10,48 @@
   }
 
   /* PRO label */
-  &.pro > .menu__list-item-collapsible > .menu__link, // With subitems
-  &.pro > .menu__link // Without subitems
-  {
+  &.pro {
     position: relative;
 
     &::after {
       content: "PRO";
+
+      // Technical settings:
+      display: block;
+      box-sizing: content-box;
       position: absolute;
+
+      // Sizing and placement:
+      line-height: 100%;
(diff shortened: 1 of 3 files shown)


Piotr1215:
Thank you @adiron. I'm still not seeing the `pro` label on Kubevirt and other standalone elements:

image

https://deploy-preview-447--vcluster-docs-site.netlify.app/docs/vcluster/next/integrations/kubevirt/

adiron:
I was very focused on the style issue and missed that bit. Interestingly, here it looks like the `sidebar_class_name` does not get applied, which is why there's no tag. I'm investigating

adiron:
Pulled Kubevirt and Cert Manager out of their folders which fixed the classes not being applied.

Piotr1215 (approved):
Thank you @adiron , the CSS changes look great.

PR #453, opened 2025-02-12
feature: adjust netlify redirects to latest

Piotr1215:
# Content Description

When users type part of url in the url search bar, the old settings would redirect them to the `v0.19` docs. Changing the redirects to use only recent version of the docs and removing redirects that are no longer there.

## Preview Link 

[What are virtual clusters?](https://deploy-preview-453--vcluster-docs-site.netlify.app/docs/)

### Note on testing
Try to type in the url search bar one of the redirects and see if it resolves to url correctly

## Internal Reference

Closes DOC-415
References DOC-419

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -10,52 +10,36 @@ command = """
   """
 
 [[redirects]]
-  from = "/docs/advanced-topics/*"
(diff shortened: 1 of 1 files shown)


adiron (approved):
New links appear to work

PR #457, opened 2025-02-15
Adds basic example at top of native sleep mode docs

cbron:
# Content Description

Adds example so native sleep mode docs look more like the Kube Virt integration page

## Internal Reference

None

### DIFF
--- a/vcluster/configure/vcluster-yaml/experimental/native-sleep-mode.mdx
+++ b/vcluster/configure/vcluster-yaml/experimental/native-sleep-mode.mdx
@@ -22,6 +22,19 @@ caveats and should not be enabled on virtual clusters that are using the current
 [external.platform](../external/platform).
 :::
 
+Not all workloads need to run all the time, and scaling them down saves time and money. With vCluster-native sleep mode you can scale down workloads based on a set schedule, or activity from users and ingress. 
+
+## Enable native sleep mode
+
+To enable vCluster-native sleep mode, use the following configuration inside your `vcluster.yaml`:
+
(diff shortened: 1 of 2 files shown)


vcluster/configure/vcluster-yaml/experimental/native-sleep-mode.mdx
```diff
@@ -22,6 +22,19 @@ caveats and should not be enabled on virtual clusters that are using the current
 [external.platform](../external/platform).
 :::
 
+Not all workloads need to run all the time, and scaling them down saves time and money. With vCluster-native sleep mode you can scale down workloads based on a set schedule, or activity from users and ingress. 
+
+## Enable Native Sleep Mode
+
```
cbron, line 28: Don't agree with the Vale check here, I think it looks good this way: https://www.vcluster.com/docs/vcluster/configure/vcluster-yaml/integrations/kubeVirt

Piotr1215: The `kubevirt` integration page was not changed since `vale` was introduced, that's why the headings are not sentence-style capitalized. We want to standardize the sentence-style capitalization rule across all the docs pages. Exceptions to this are projects names, acronyms etc. Here is [more info](https://developers.google.com/style/capitalization) about the capitalization rules and reasoning why we use them.

Piotr1215 (changes requested):
Can you add titles to the code snippets, something like

\```yaml title="Native sleep mode configuraiton"

PR #476, opened 2025-02-25
fix: Add UI and CLI options for share

joaniekube:
# Content Description

Makes the Ui procedure into an partial and adds CLI step

## Preview Link 

[Link to preview](https://deploy-preview-476--vcluster-docs-site.netlify.app/docs/platform/use-platform/virtual-clusters/manage-access/?x0=2)

## Internal Reference

Closes DOC-222

### DIFF
--- a/platform/_partials/vcluster/share-cli.mdx
+++ b/platform/_partials/vcluster/share-cli.mdx
@@ -1,4 +1,6 @@
 To grant a user access to a virtual cluster with Loft CLI, run the following command:
 ```bash
-loft share vcluster [optional:name] --user other-user --project my-project
+vcluster platform share vcluster VCLUSTER_NAME [flags]
 ```
+
+See the [Loft CLI documentation](/vcluster/cli/vcluster_platform_share_vcluster) for more information.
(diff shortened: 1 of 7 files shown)


joaniekube:
These sections need to be rewritten. @Piotr1215 May I address the vale comments when I work on these docs. I notice the headings did not use sentence case so fixed those, but I will need more time to rewrite the sections, and the content edits are out of scope for this issue.

platform/_partials/vcluster/share-cli.mdx
```diff
@@ -0,0 +1,4 @@
+To grant a user access to a virtual cluster with Loft CLI, run the following command:
+```bash
+loft share vcluster [optional:name] --user other-user --project my-project
```
Piotr1215, line 3: This command is for the older version of vCluster `v0.19`. In versions `v0.20+` the command should be:

`vcluster platform share vcluster VCLUSTER_NAME [flags]`

Here is the reference: https://www.vcluster.com/docs/vcluster/cli/vcluster_platform_share_vcluster

Piotr1215 (changes requested):
Don't worry about the vale issues, we can address those separately.

platform/_partials/vcluster/share-cli.mdx
joaniekube: I updated this change, but should I also update the versioned docs (e.g. `0.23.0`) ?

PR #485, opened 2025-03-03
fix: add UK spelling

joaniekube:
# Content Description

- Enable vale spell checker and add UK spelling as words to look out for.


joaniekube:
PTAL @Piotr1215. Does this look okay?

Piotr1215 (changes requested):
Thank you, @joaniekube , we should mention that this comes from GitLab ruleset, just reversed.

The substitution rule doesn't have keyword `pattern`, you can validate `vale` patterns in https://studio.vale.sh/.

joaniekube:
removed the regex @Piotr1215 and added more words to the list

Piotr1215 (approved):

PR #490, opened 2025-03-05
chore(vale): add missing American spelling words

joaniekube:
# Content Description

- Add spellings to vale list
    - fantasise vs. fantasize
    - idolise vs. idolize

### DIFF
--- a/.github/styles/Google/AmericanSpelling.yml
+++ b/.github/styles/Google/AmericanSpelling.yml
@@ -43,6 +43,7 @@ swap:
   enrol: enroll
   enrolment: enrollment
   enthral: enthrall
+  fantasise: fantasize
   favourite: favorite
   fibre: fiber
   fillet: filet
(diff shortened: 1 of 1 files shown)


Piotr1215 (changes requested):
See if this rule would work for covering all the `ise -> ize` and `ism -> izm`

```yaml
extends: substitution
message: "Use '%s' instead of '%s'."
level: error
ignorecase: false
swap:
  '\b(\w+)ism\b': '$1izm'  # Converts words ending in "ism" to "izm"
  '\b(\w+)ise\b': '$1ize'  # Converts words ending in "ise" to "ize"
```

joaniekube:
Hey @Piotr1215, I think it would be more effective to simply list the words to avoid rather than creating a general rule with exceptions. Most of the words in our list are unlikely to appear in the docs, and only a few (like apologize, organization, standardize) might cause issues for contributors. Defining an absolute rule like  ise -> ize and ism -> izm and then listing exceptions will most likely make the linter really noisy and create more issues for us. WDYT?

Piotr1215:
I'm ok with the list.

Piotr1215 (approved):

PR #500, opened 2025-03-07
feat: add kb article content about difference in allocatable value

Ilan-loft:
KB article about difference in allocatable value betwwen vcLuster and host

# Content Description

## Internal Reference

Closes DOC-


joaniekube:
Hi @Ilan-loft . Thanks for opening this PR! I'm trying to determine where this fits—perhaps it doesn’t need to be a standalone doc. I’d like to better understand its impact from a user perspective. I’ll ping you on Slack, to possibly set something up.

joaniekube (approved):
LGTM

PR #501, opened 2025-03-10
docs: improve snapshot & restore docs

FabianKramm:


Piotr1215 (approved):
We can merge it into the `main` docs version so that users trying out the new alpha/rc releases can have some documentation.

PR #503, opened 2025-03-10
Update HA and PVC incompatibility to make it more obvious

deniseschannon:
# Content Description

For audit logging persistence, it was hidden that PVC method was not supported, so made it a call out. 

## Internal Reference

Closes ENG-3792


joaniekube (approved):
LGTM

PR #511, opened 2025-03-12
feat: create airgapped vcluster deployment docs

prakhar-loft:
Deploy vcluster in an air-gaped env without an internal helm repo

# Content Description

## Internal Reference

Closes DOC-


joaniekube:
I think this information might fit better in another section/ or in install docs that already exist. I’ll continue reviewing and suggest some edits, but we may be able to consolidate it with the existing installation docs. I need to look into it a little further.

PR #533, opened 2025-03-18
fix: update to include only 1.30 and higher

joaniekube:
# Content Description

- Update table to reflect support for 1.30 upwards.

## Internal Reference

Closes DOC-511


joaniekube:
I think this is what we had in the previous version @Piotr1215 

image

Piotr1215 (approved):

PR #550, opened 2025-03-21
fix: correct typos

joaniekube:
# Content Description

- Correct typos highlighted in `0.24.0` release

## Internal Reference

Closes DOC-517


Piotr1215 (approved):
Thank you @joaniekube , this will be helpful when cutting a new release. We will need to follow up and change non-versioned partials references to @site , I'll create issue for it.

PR #583, opened 2025-04-02
fix: reference correct version in ESO example

johannesfrey:
# Content Description
The ESO example was referencing a wrong version number.

## Internal Reference

part of ENG-6173


joaniekube (approved):
LGTM

PR #592, opened 2025-04-07
fix: add manual namespace labels

joaniekube:
# Content Description

- Add manual labels needed for namespace creation

## Internal Reference

Closes DOC-77


Piotr1215 (approved):

Piotr1215 (changes requested):
Need to place it at the start, maybe first thing in the # Deployment section.

PR #601, opened 2025-04-10
Create helm-permission-denied.mdx

prakhar-loft:
helm permission denied error while creating the vCluster using vCluster-platform.

# Content Description

## Internal Reference

Closes DOC-


kmadel:
Looks good to me. The only caveat might be if Pod Security Standards are enabled on the host with the restricted profile. I believe that even `/tmp` is blocked in that case.

Piotr1215 (approved):

PR #640, opened 2025-04-28
Add istio example to sleep mode docs

zerbitx:
# Content Description

## Preview Link 

[Istio Example Preview](https://deploy-preview-640--vcluster-docs-site.netlify.app/docs/vcluster/next/configure/vcluster-yaml/sleep-mode#istio-gateway)

## Internal Reference

Closes ENG-6495

### DIFF
--- a/vcluster/_fragments/sleepmode-istio-example.mdx
+++ b/vcluster/_fragments/sleepmode-istio-example.mdx
@@ -9,6 +9,9 @@ import SleepmodeIstioExample from '!!raw-loader!@site/vcluster/configure/vcluste
   <Step>
 Create the `kind` cluster.
 
+Here we are pinning the ports 80, 443 and the status port 15021 to specific ports in the Kubernetes node port range.  These
+ are used by Istio Gateway and needed for running this example with `kind`.
+
 ```shell title="Create a kind cluster"
 kind create cluster --name istio-demo --config - <<EOF
 kind: Cluster
@@ -43,11 +46,10 @@ EOF
     ```shell title="Install with ambient"
     istioctl install --set profile=ambient --set "components.ingressGateways[0].enabled=true" --set "components.ingressGateways[0].name=istio-ingressgateway" --skip-confirmation
     ```
-
   </Step>
 
   <Step>
-Patch the Istio gateway's service to match the ports in the kind configuration.
+The Istio Gateway service starts with some random node ports.  To line them up with the kind cluster configuration, patch the Istio gateway service.
 
 ```shell title="Patch ingress service ports"
 kubectl patch service istio-ingressgateway -n istio-system --patch "$(cat <<EOF
@@ -72,7 +74,7 @@ EOF
   <Step>
     Create the vCluster.
 
-    Use the following `vcluster.yaml` to create a virtual cluster on your host. Save this file as `vcluster.yaml`
+    Use the following `vcluster.yaml` to create a virtual cluster on your host cluster. Save this file as `vcluster.yaml`
 
     <CodeBlock title="vCluster config for auto sleep" language="yaml">{SleepmodeIstioConfig}</CodeBlock>
     And run the following command:
--- a/vcluster/_fragments/sleepmode-istio-example.mdx
+++ b/vcluster/_fragments/sleepmode-istio-example.mdx
@@ -1,0 +1,109 @@
+import Flow, { Step } from '@site/src/components/Flow';
+import CodeBlock from '@theme/CodeBlock'
+import SleepmodeIstioConfig from '!!raw-loader!@site/vcluster/configure/vcluster-yaml/_code/sleepmode-istio-config.yaml'
+import SleepmodeIstioExample from '!!raw-loader!@site/vcluster/configure/vcluster-yaml/experimental/_code/sleepmode-istio-examples.yaml'
+
+### Configure sleep mode for use with Istio
+
+<Flow id="sleepmode-ingress-example">
+  <Step>
+Create the `kind` cluster.
+
+```shell title="Create a kind cluster"
+kind create cluster --name istio-demo --config - <<EOF
+    kind: Cluster
+    apiVersion: kind.x-k8s.io/v1alpha4
+    nodes:
+    - role: control-plane
+    extraPortMappings:
+    - containerPort: 30000
+    listenAddress: 127.0.0.1
+    hostPort: 80
+    protocol: TCP
+    - containerPort: 31000
+    listenAddress: 127.0.0.1
+    hostPort: 443
+    protocol: TCP
+    - containerPort: 32000
+    listenAddress: 127.0.0.1
+    hostPort: 15021
+    protocol: TCP
+EOF
+```
+  </Step>
+  <Step>
+    Install istio
+
+    ```shell title="Install istio"
+    istioctl install -y
+    ```
+  </Step>
+
+  <Step>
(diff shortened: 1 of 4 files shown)


vcluster/_fragments/sleepmode-istio-example.mdx
```diff
@@ -0,0 +1,116 @@
...
+import SleepmodeIstioExample from '!!raw-loader!@site/vcluster/configure/vcluster-yaml/experimental/_code/sleepmode-istio-examples.yaml'
+
+### Configure sleep mode for use with Istio
+
+<Flow id="sleepmode-ingress-example">
+  <Step>
+Create the `kind` cluster.
+
+```shell title="Create a kind cluster"
+kind create cluster --name istio-demo --config - <<EOF
```
cbron, line 13: How much time would it take to switch this from kind to a real cluster ? Or have both side by side. 
If it takes a while, for sake of timing, we could merge this one and do that later on.

vcluster/_fragments/sleepmode-istio-example.mdx
```diff
@@ -0,0 +1,116 @@
...
+    extraPortMappings:
+      - containerPort: 30000
+        listenAddress: 127.0.0.1
+        hostPort: 80
+        protocol: TCP
+      - containerPort: 31000
+        listenAddress: 127.0.0.1
+        hostPort: 443
+        protocol: TCP
+      - containerPort: 32000
```
cbron, line 27: Where do these ports come from, are they random just for this setup ?

vcluster/_fragments/sleepmode-istio-example.mdx
```diff
@@ -0,0 +1,116 @@
...
+        listenAddress: 127.0.0.1
+        hostPort: 15021
```
cbron, line 29: Same question here

vcluster/_fragments/sleepmode-istio-example.mdx
```diff
@@ -0,0 +1,116 @@
...
+    Install istio
+
+    ```shell title="Install istio"
+    istioctl install -y
+    ```
+
+    or
+
+    ```shell title="Install with ambient"
+    istioctl install --set profile=ambient --set "components.ingressGateways[0].enabled=true" --set "components.ingressGateways[0].name=istio-ingressgateway" --skip-confirmation
```
cbron, line 44: How come only ambient has ingressGateway turned on ? Wouldn't it be on for both ambient and not

vcluster/_fragments/sleepmode-istio-example.mdx
```diff
@@ -0,0 +1,116 @@
...
+    nodePort: 32000
+EOF
+)"
+```
+  </Step>
+
+  <Step>
+    Create the vCluster.
+
+    Use the following `vcluster.yaml` to create a virtual cluster on your host. Save this file as `vcluster.yaml`
```
cbron, line 75:
```suggestion
    Use the following `vcluster.yaml` to create a virtual cluster on your host cluster. Save this file as `vcluster.yaml`
```

vcluster/configure/vcluster-yaml/sleep-mode.mdx
```diff
@@ -187,6 +188,17 @@ To configure sleep mode for the ingress controller, make sure it stays responsiv
 <SleepModeIngressExample />
 </details>
 
+### Istio gateway
+
+With Istio installed on the host cluster and the [Istio integration](./integrations/istio) enabled, [Gateway](https://istio.io/latest/docs/reference/config/networking/gateway/) and [VirtualService](https://istio.io/latest/docs/reference/config/networking/virtual-service/) resources created in the vCluster
```
cbron, line 193: This implies that the integration is required for this to work, is that true ? If so we should clarify.
If it's not true, is there anything we have to do to turn it on with vanilla istio ? Or is the sleep controller smart enough to figure out ingress-nginx vs istio ingress.

cbron (changes requested):

vcluster/_fragments/sleepmode-istio-example.mdx
zerbitx: Random(ish). The Istio install will choose random container ports.  Setting them explicitly here lets us patch the istio service when it starts to be able to hit port 80 from the host with curl. I can add an explanation.

vcluster/_fragments/sleepmode-istio-example.mdx
zerbitx: It's simpler on a real cluster, or at least on OrbStack where I did most of my dev/testing.  I wrote the example in kind to make clean up easier.

vcluster/_fragments/sleepmode-istio-example.mdx
zerbitx: A result of trying different things to get things working with kind and landing on the first one that worked.

vcluster/configure/vcluster-yaml/sleep-mode.mdx
zerbitx: It is true, so I'll call that out more explicitly.  WIthout the integration enabled, the sleep mode controller isn't started.  Otherwise it would be making modifications to Istio resources in the virtual cluster that never get synced/translated to the host.

vcluster/_fragments/sleepmode-istio-example.mdx
cbron: Ok, can you just clarify that in this. Like: "Istio will have specific ports, set them here..."

vcluster/_fragments/sleepmode-istio-example.mdx
cbron: haha, ok, but then are you saying that the first one (sidecar) worked without these flags ?

vcluster/_fragments/sleepmode-istio-example.mdx
zerbitx: Scratch that, at least if we want the instructions to deal with ambient mode specifically, in the interest of time we should leave it with kind.

cbron: Ok, can you work on that later this week. Feel free to open up new issue.

vcluster/_fragments/sleepmode-istio-example.mdx
zerbitx: Correct.  If I just do the standard install sleeping and waking work fine.

cbron: If it doesn't matter, wouldn't it also make sense to add it to one above too ? That way it would work all the way from outside of cluster for both options.

cbron (approved):
1 thing left, but approving now so that we can fast-track due to timelines. We can address that thing at later date, it's not critical.

vcluster/_fragments/sleepmode-istio-example.mdx
```diff
@@ -0,0 +1,118 @@
...
+        protocol: TCP
+EOF
+```
+  </Step>
+  <Step>
+    Install istio
+
+    ```shell title="Install istio"
+    istioctl install -y
+    ```
```
cbron, line 42: Should remove since we don't support non-ambient. I've resolved other convo related to this.

joaniekube:
@zerbitx which is the preferred PR to merge. This one or #640?

PR #649, opened 2025-04-30
update: Index page for cost saving feature

hrittikhere:
@Piotr1215 I’ve added the PR for the documentation with more details, which should help reduce redundancy. When you get a chance, let me know your thoughts.


Piotr1215:
@hrittikhere looks like it breaks several styling issues. What is the reasoning behind adding this? It looks like a significant change in a highly visible document which needs careful wording and styling. Can you open a linear issue instead where we can discuss what is required?

PR #688, opened 2025-05-14
document what's translated and not supported in the istio CRs

hidalgopl:
# Content Description
added a note about what's being translated in the sync of istio resources and what is not supported.

## Internal Reference
Closes ENG-6861


joaniekube (approved):
LGTM

PR #709, opened 2025-05-21
fix: node IP visibility

sydorovdmytro:
# Content Description
Removed some example commands that don't have any effect or may lead to issues.
Node IPs for the virtual cluster are always obfuscated and changing this behavior is not possible or tricky.


joaniekube (approved):
LGTM

PR #713, opened 2025-05-22
fix: update sleep mode docs for platform

joaniekube:
# Content Description

## Internal Reference

Closes DOC-522
Closes DOC-471


joaniekube:
@cbron I need some guidance on this one. Do you want to completely remove the platform docs  for sleep mode, or should we replace it with the sleep mode content that is in vCluster?

zerbitx:
> @cbron @zerbitx I need some guidance on this one. Do you want to completely remove the platform docs for sleep mode, or should we replace it with the sleep mode content that is in vCluster?

I'm not sure the best way to organize this since they are two different features, essentially, with a unified way to configure the "sleep" portion.  There's no auto deletion without the platform and activity monitoring is different between the two.  There should still be documentation in both places, but I don't know where we'd like the bulk of the talk about what it is, benefits and examples.  I think we want as many people as possible on platform, so maybe we should be removing all but the config reference on the vCluster side and trying to sell/explain it only on the platform side?

joaniekube:
Hi @zerbitx 

Thanks for laying this out and providing more context. 

One thing to consider is where our users are coming in from - Are most users first encountering this feature while using vCluster directly, and then exploring platform capabilities later? Or are we primarily guiding users from the Platform side first?

If the typical entry point is vCluster, I agree with what you commented - explanation in the vCluster docs — enough to explain what sleep mode is, that it can be extended on the Platform, and provide the basic config reference. 

We can then use the Platform docs to go deeper, highlighting auto-deletion, activity monitoring, and the benefits of upgrading. 

So maybe:

vCluster docs = intro + config + activity monitoring +  link to Platform
Platform docs = full breakdown  (including activity monitoring for platform) + use cases + examples

WDYT?

Piotr1215:
@zerbitx we would like to merge this issue tomorrow before platform docs release, does the content look ok technically from your side?

PR #732, opened 2025-05-29
fix Update air-gapped doc to remove mention of license key

joaniekube:
# Content Description

## Internal Reference

Closes DOC-661


joaniekube:
Working on this. Will complete/merge Monday

PR #740, opened 2025-05-30
fix(infra): javascript calls overload in dev mode

Piotr1215:
# Content Description

- Fixed React warnings by changing `class` to `className` in pro-admonition.mdx partials across all versions to prevent errors like these:

Screenshot from 2025-05-30 13-59-21

- Removed infinite recursion in details-clicks.js that was causing "Too many calls to Location or History APIs" errors
- Added proper ExecutionEnvironment checks for browser-specific functions to prevent SSR issues
- Tested on production build and confirmed no regression errors

Browser errors like these should no longer appear when running `npm run start` on configuration pages with pro/enterprise labels.

image

## Preview Link 

https://deploy-preview-740--vcluster-docs-site.netlify.app/docs/vcluster/next/configure/vcluster-yaml/

## Internal Reference

Closes DOC-665

PR #784, opened 2025-06-10
fix(vcluster): re-enable copy button

Piotr1215:
# Content Description

This is the regression error introduced in #740

Restore copy button functionality by adding targeted event listeners for `DOMContentLoaded`, `hashchange` (sidebar navigation), `visibilitychange` (tab switching), and page change detection using MutationObserver that watches for URL changes during Docusaurus navigation. 

These calls only execute when config fields are present, avoiding the aggressive polling that caused "Too many calls to Location or History APIs" while ensuring copy buttons appear reliably on all config pages.

## Internal Reference

Closes DOC-703

### DIFF
--- a/src/js/details-clicks.js
+++ b/src/js/details-clicks.js
@@ -608,12 +608,47 @@ const preserveExpansionStates = ExecutionEnvironment.canUseDOM ? function(skipEv
     }
   });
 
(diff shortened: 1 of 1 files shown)


matskiv (dismissed):
That seems to work. Thank you!

PR #786, opened 2025-06-10
fix: Make private nodes not visible on main

joaniekube:
# Content Description

- Remove private nodes from main docs

## Internal Reference

Closes DOC-705


Piotr1215 (approved):
`next` and `main` are now synced, from this point we are tracking the private nodes deployment docs in `next` branch.
The liner action failed due to [linear API outage](https://linearstatus.com/).

PR #807, opened 2025-06-16
Since 0.25 it is not possible to have multiple vCluster instances in a single host namespace

kmadel:
Removed sentence that falsely states running multiple vCluster instances in the same host namespace is possible.

# Content Description

## Internal Reference

Closes DOC-


joaniekube (approved):
lgtm

joaniekube (approved):

PR #840, opened 2025-06-30
refactor: updates for templates

joaniekube:
# Content Description

## Preview Link 

[Link to Preview](https://deploy-preview-840--vcluster-docs-site.netlify.app/docs/platform/next/administer/templates/overview)

## Internal Reference

Closes DOC-719

### DIFF
--- a/platform/administer/templates/create-templates.mdx
+++ b/platform/administer/templates/create-templates.mdx
@@ -19,7 +19,9 @@ Templates standardize virtual cluster, namespace, and app configurations. Create
 
 ## Create a template using the UI
 
-Templates can be created directly in the UI.
+You can create templates directly in the UI:
+
+<br/>
 
 <Tabs>
   <TabItem value="virtual-cluster-template" label="Virtual cluster template" default>
@@ -218,20 +220,18 @@ spec:
               enabled: true
 ```
 
-## Configure template options {#template-configuration-options}
-
-Templates include configuration sections that control template behavior, content, and access. The following configuration options can be applied to virtual cluster templates and namespace templates.
-
 
 ## Configure template options {#template-configuration-options}
 
 Templates include configuration sections that define behavior, content, and access. The following options apply to both virtual cluster templates and namespace templates.
+
 
 ### Template parameters
 
 Template parameters make templates dynamic by exposing configurable fields that users can customize during deployment. When creating parameters, you define the variable name, data type, label, and description.
 
 Parameters can include validation rules such as allowed options, required fields, and default values. For example, you might create parameters for admin passwords, environment types, resource limits, or feature toggles. Users see these parameters as form fields. They can customize deployments without modifying the underlying template.
+
 
 ### Apps and objects
 
@@ -240,6 +240,7 @@ The **Apps** section lets you select pre-built applications that install automat
 Selected apps deploy immediately after virtual cluster creation. Development teams have required tools and services available without manual installation. This approach standardizes the application stack across virtual clusters. It also reduces setup time for development teams.
 
 The **Objects** section defines Kubernetes manifests that are applied during virtual cluster creation. These include ConfigMaps, Secrets, or NetworkPolicies that establish baseline configurations.
+
 
 ### Management access
 


platform/administer/templates/overview.mdx
```diff
@@ -0,0 +1,820 @@
...
+**Compare template versions:**
+
+```bash
+# View differences between template versions
+loft diff template development-template --from-version 1.2.2 --to-version 1.2.3
+```
+
+### Version rollback scenarios
+
+Template versioning becomes crucial when you need to rollback problematic changes:
```
lizardruss, line 439: This section might be confusing. I've added corrections to the commands, but they would only apply to each vCluster. If the goal is to modify multiple vCluster to use a new template version, that's only possible using "wild card" versions. If a VirtualClusterInstance refers to a template version with an `x` (ex: `1.0.x`, it will automatically upgrade when a new version (like `1.0.1`) is created.

platform/administer/templates/overview.mdx
```diff
@@ -0,0 +1,820 @@
...
+For complex business logic validation, you can define custom validation functions:
+
+```yaml
+spec:
+  parameters:
+    - name: cluster_size
+      type: string
+      validation:
+        # Custom validation function
+        customValidation: |
```
lizardruss, line 788: I don't think this exists. There's support for `validation` and `invalidation` regexes.

joaniekube:
@kmadel provided some additional examples that have been helpful to customers, so adding that to the overview page as well

platform/administer/templates/index.mdx
```diff
@@ -0,0 +1,261 @@
...
+import Flow, { Step } from "@site/src/components/Flow";
+import Button from "@site/src/components/Button";
+import Label from "@site/src/components/Label";
+import NavStep from "@site/src/components/NavStep";
+import PartialVirtualClusterCreateOptionalTemplateUI from "../../_partials/vcluster/create-ui-optional-fields-template.mdx";
+
+
+# Templates
+
+vCluster templates define reusable configurations for creating virtual Kubernetes clusters. They allow platform teams to standardize cluster creation while giving users a controlled way to customize certain settings. Templates are implemented using Kubernetes Custom Resources (`VirtualClusterTemplate`) and can be managed with tools like `kubectl`, Helm, or Terraform.
```
kmadel, line 18: I think I would mention Argo CD (especially since we have an integration) and not sure about Terraform as it isn't K8s native like Argo CD.

platform/administer/templates/index.mdx
```diff
@@ -0,0 +1,261 @@
...
+
+Each template includes the base configuration for a virtual cluster—such as the Kubernetes version, control plane type, networking mode, and storage settings. Templates can also define resource constraints (CPU, memory, storage) to prevent overconsumption and ensure predictable usage.
```
kmadel, line 20: Probably should add sync settings as well, as that is one of the most important configurations of a vCluster.

deniseschannon:
Add a section around Delete virtual cluster and the implications around it, where deleting a template that a virtual cluster references will not break the virtual cluster, but it makes the virtual cluster stuck in that state forever. You will not be able to edit/upgrade that virtual cluster as the reference of that template is gone.

platform/administer/templates/index.mdx
```diff
@@ -0,0 +1,55 @@
...
+| Virtual cluster              | `VirtualClusterTemplate`                                  | Specifies how virtual clusters are provisioned, including control plane settings, synchronization behavior, and metadata. | Provision virtual clusters with consistent architecture, features, and default settings. |
+| App                          | `AppTemplate`                                             | Defines reusable application configurations, such as Helm charts or container specifications.                             | Deploy standard applications across projects or spaces using consistent settings.        |
+| Namespace                    | `NamespaceTemplate`                                       | Predefines configurations for new namespaces, including role-based access control, resource quotas, and labels.           | Automatically apply baseline policies when a namespace is created.                       |
```
lizardruss, line 36: The custom resource for namespaces is `SpaceTemplate` ([ref](https://github.com/loft-sh/loft-enterprise/blob/main/staging/src/github.com/loft-sh/api/v4/pkg/apis/management/v1/spacetemplate_types.go#L15))

platform/administer/templates/parameters.mdx
```diff
@@ -0,0 +1,422 @@
...
+
+Templates support basic parameter types for collecting user input during cluster creation:
+
+| Type          | Use case                                               |
+|---------------|--------------------------------------------------------|
+| [`String`](#string-parameter)    | Names, labels, image tags, configuration values        |
+| [`Number`](#string-parameter)    | Replica counts, port numbers, resource limits          |
+| [`Boolean`](#boolean-parameter)  | Feature toggles, enable/disable flags                  |
+| [`Password`](#password-parameter)| API tokens, secrets, sensitive input values            |
+| [`Multiline`](#multiline-parameter)| Configuration files, startup scripts, long descriptions |
```
lizardruss, line 36: Unsure if its a docs convention to uppercase these, but it might be better to lowercase them. That way it matches the casing in [the examples](https://deploy-preview-840--vcluster-docs-site.netlify.app/docs/platform/next/administer/templates/parameters/#number-parameter) and they are not tempted to capitalize it in config. I'm not sure if the backend will accept the following, for example:
```yaml
- variable: replica_count
  type: Number # uppercase instead of lowercase
```

platform/administer/templates/parameters.mdx
```diff
@@ -0,0 +1,422 @@
...
+- **Use descriptive names**: Choose parameter identifiers that clearly communicate their purpose.
+- **Write clear descriptions**: Include concise guidance, examples, or constraints in the parameter description field to help users enter correct values.
+
+You can also apply advanced parameter options to modify behavior and enforce consistency:
+
+- **Default values**: Use defaults to reduce manual input and guide users toward standard configurations. For example, you can set a default replica count or enable a common add-on by default.
+
+- **Conditional display**: Show parameters only when they are relevant. For example, display a domain name field only if ingress is enabled.
+
+- **Dependencies**: Link parameters so that one selection controls another. For example, choosing a cloud provider can change the list of available regions.
```
lizardruss, line 100: I don't think these are supported today. ([ref](https://github.com/loft-sh/loft-enterprise/blob/main/staging/src/github.com/loft-sh/api/v4/pkg/apis/storage/v1/app_types.go#L136-L189))

platform/administer/templates/parameters.mdx
```diff
@@ -0,0 +1,422 @@
...
+
+```yaml title="Number parameter with range"
+- variable: replica_count
+  label: Replica Count
+  description: Number of application replicas (1-10)
+  type: number
+  required: true
+  min: 1
+  max: 10
+  default: 3
```
lizardruss, line 147:
```suggestion
  defaultValue: 3
```

platform/administer/templates/parameters.mdx
```diff
@@ -0,0 +1,422 @@
...
+```
+
+#### Boolean parameter for feature toggle {#boolean-parameter}
+
+```yaml title="Boolean parameter for feature toggle"
+- variable: enable_monitoring
+  label: Enable Monitoring
+  description: Install Prometheus and Grafana monitoring stack
+  type: boolean
+  default: true
```
lizardruss, line 158:
```suggestion
  defaultValue: true
```

platform/administer/templates/parameters.mdx
```diff
@@ -0,0 +1,422 @@
...
+
+- variable: storage_size
+  label: Storage Size
+  description: Size of persistent storage (e.g., 10Gi, 50Gi)
+  type: string
+  required: true
+  validation: "^[0-9]+[KMGT]i$"
+  condition: "{{ .Values.enable_persistence }}"
+  section: Storage Settings
+```
```
lizardruss, line 229: Remove since I don't think `condition` is supported. @kmadel Might know for sure, but I don't see support for this in the backend.

platform/administer/templates/parameters.mdx
```diff
@@ -0,0 +1,422 @@
...
+## Access parameter values in templates
+
+After you define parameters, you can access their values in your template using Go template syntax. The platform makes all parameter values available under the `.Values` object.
+
+### Basic parameter access
+
+```yaml
+apiVersion: vcluster.loft.sh/v1
+kind: VCluster
+metadata:
```
lizardruss, line 240: Not a real resource in vCluster platform

platform/administer/templates/parameters.mdx
```diff
@@ -0,0 +1,422 @@
...
+    grafana:
+      enabled: true
+  {{- end }}
+  
+  {{- if .Values.enable_persistence }}
+  storage:
+    persistence: true
+    size: "{{ .Values.storage_size }}"
+  {{- end }}
+```
```
lizardruss, line 292: I think this is intended to be `vcluster.yaml`, but it doesn't line up with the spec.

platform/administer/templates/parameters.mdx
```diff
@@ -0,0 +1,422 @@
...
+        # Ensure resource values are valid
+        cpu: >-
+          {{- if .Values.cpu_limit -}}
+          {{ .Values.cpu_limit }}
+          {{- else if eq .Values.environment_type "production" -}}
+          2000m
+          {{- else -}}
+          1000m
+          {{- end }}
+```
```
lizardruss, line 422: Similar comment as above. If this is intended to be a generic templating example, it resembles `vcluster.yaml` closely without matching the spec and could cause confusion

platform/administer/templates/parameters.mdx
```diff
@@ -0,0 +1,422 @@
...
+metadata:
+  name: "{{ .Values.cluster_name | lower }}"
+  labels:
+    # Use default value if parameter is empty
+    cost-center: "{{ .Values.cost_center | default "engineering" }}"
+    # Combine multiple parameters
+    full-name: "{{ .Values.project_name }}-{{ .Values.environment_type }}"
+    # Transform string values
+    normalized-name: "{{ .Values.cluster_name | lower | replace "_" "-" }}"
+```
```
lizardruss, line 308: Would it be possible to provide a list of available functions? This mentions `replace`, `lower`, and `default`, so naturally I'd be curious what else is available.

FWIW, I think we support everything here except `env` and `envexpand`: https://masterminds.github.io/sprig/

We also add in several other functions. Their names and descriptions can be found here: https://github.com/loft-sh/loft-enterprise/blob/main/pkg/helm/engine/funcs.go#L47-L55

lizardruss (changes requested):
Added some more feedback, mostly what is/is not supported by the backend.

platform/administer/templates/overview.mdx
```diff
@@ -0,0 +1,55 @@
...
+import TabItem from '@theme/TabItem';
+import Flow, { Step } from "@site/src/components/Flow";
+import Button from "@site/src/components/Button";
+import Label from "@site/src/components/Label";
+import NavStep from "@site/src/components/NavStep";
+
+
+# Overview
+
+vCluster templates define reusable configurations for creating virtual Kubernetes clusters. They allow platform teams to standardize cluster creation while giving users a controlled way to customize certain settings. Templates are implemented using Kubernetes custom resources (`VirtualClusterTemplate`) and can be managed with tools such as `kubectl`, Helm, or ArgoCD.
```
kmadel, line 17: Is it intentional to use ‘vCluster’ here instead of ‘virtual cluster templates’?

Also, in the second sentence should probably add ‘virtual’ - so  ‘… standardize virtual cluster creation …’

platform/administer/templates/overview.mdx
```diff
@@ -0,0 +1,55 @@
...
+
+Each template defines the base configuration for a virtual cluster, including the Kubernetes version, control plane type, networking mode, storage settings, and sync behavior. Sync settings control which Kubernetes resources are synchronized between the virtual cluster and the host cluster—such as Service, Secret, ConfigMap, or custom resources. Templates can use sync rules to import host resources, export virtual cluster resources, or customize synchronization at the namespace or cluster scope.
```
kmadel, line 19: What is ‘control plane type’ and ‘network mode’ referring to?
Not sure I would use `Services as a sync example since they are synced by default - `Ingress` would be a good, common example.

Not sure I follow “Templates can use sync rules to import host resources, export virtual cluster resources, or customize synchronization at the namespace or cluster scope.” You can do this without a template with a regular vcluster.yaml, not sure it is necessary here as part of virtual cluster template documentation.

platform/administer/templates/overview.mdx
```diff
@@ -0,0 +1,55 @@
...
+
+To support secure and consistent self-service, templates allow administrators to define which settings users can modify through parameters. For example, you can expose only a few fields—such as the number of replicas or a set of approved storage classes—while keeping the rest of the configuration fixed. This helps reduce the risk of misconfiguration, but also enables flexibility.
+
+Templates can also include preinstalled apps, such as monitoring agents or ingress controllers, using Helm charts or Kubernetes manifests. Additional Kubernetes resources like ConfigMaps, RBAC, or NetworkPolicies can be applied to either the host cluster or the virtual cluster itself.
+
+Because templates can be versioned, they support upgrade workflows. When a template is updated, those changes can be propagated to any virtual clusters based on it. This allows teams to update cluster configurations and installed apps consistently across environments.
```
kmadel, line 25: Maybe update to read “ those changes can be manually or automatically propagated”

Suggestion, “update cluster configurations and installed apps consistently across a fleet of virtual clusters”

platform/administer/templates/overview.mdx
```diff
@@ -0,0 +1,55 @@
...
+| ---------------------------- | --------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------- | ---------------------------------------------------------------------------------------- |
+| Virtual cluster              | `VirtualClusterTemplate`                                  | Specifies how virtual clusters are provisioned, including control plane settings, synchronization behavior, and metadata. | Provision virtual clusters with consistent architecture, features, and default settings. |
+| App                          | `AppTemplate`                                             | Defines reusable application configurations, such as Helm charts or container specifications.                             | Deploy standard applications across projects or spaces using consistent settings.        |
```
kmadel, line 35: Shouldn’t “container specifications” be “Kubernetes manifests”

platform/administer/templates/overview.mdx
```diff
@@ -0,0 +1,55 @@
...
+You can [create templates](/platform/next/administer/templates/create-templates) using the vCluster Platform UI for guided configuration or by applying manifests directly with kubectl, Helm, or ArgoCD. 
+
+
+## Apply parameters and versioning
+
+Templates support optional parameterization and versioning. Optionally, you can apply 
+
+- [Parameters](/platform/next/administer/templates/parameters): Enable dynamic customization of template fields, allowing users to specify values for CPU limits, labels, environment types, and other configuration options during vCluster provisioning.
+
+- [Versioning](/platform/next/administer/templates/versioning): Enables you to manage and iterate template configurations across different releases, providing version control for your cluster templates.
```
kmadel, line 48: Versioning also allows for optional automatic update of virtual clusters based on those versioned templates. Non-versioned templates do not support automatic update of virtual clusters.

platform/administer/templates/versioning.mdx
```diff
@@ -0,0 +1,103 @@
...
+sidebar_label: Versioning
+sidebar_position: 4
+---
+
+import Flow, { Step } from "@site/src/components/Flow";
+import NavStep from "@site/src/components/NavStep";
+import Button from "@site/src/components/Button";
+import Label from "@site/src/components/Label";
+
+Templates can be versioned. Versioning a template allows you
```
kmadel, line 12: Does this need “Templates can be versioned.” I think you could just start with the second sentence.

platform/administer/templates/versioning.mdx
```diff
@@ -0,0 +1,103 @@
...
+to make changes over time to a template while retaining the current version/iteration. Resources (for example virtual clusters, namespaces) that are deployed with templates that have updated versions show up with a warning
+indicating that a newer template version is available.
+
+## Template version strings
+
+When creating a virtual cluster from a template, either through the UI or CLI, you can specify a template version string. If omitted, the latest available version is selected by default.
```
kmadel, line 18: I would remove “ either through the UI or CLI,” since you leave out other ways (`VirtualClusterInstance` CRD via kubectl or any other K8s tooling) and doesn’t add much here.

platform/administer/templates/versioning.mdx
```diff
@@ -0,0 +1,103 @@
...
+
+You can use the `X` wildcard in the major, minor, or patch components of the version string to enable automatic upgrades. This allows virtual clusters to automatically update to newer compatible versions of the template.
+
+For example given a template version `1.0.0`:
+
+- If you set the version string to `1.X.X`, and a new template version `1.1.0` is added, the virtual cluster will automatically upgrade to `1.1.0`.
+- If a version `2.0.0` is added instead, the cluster **will not** upgrade, because the major version does not match.
```
kmadel, line 25: “ the cluster” to “ the virtual cluster”

platform/administer/templates/versioning.mdx
```diff
@@ -0,0 +1,103 @@
...
+
+To always upgrade to the latest available version, you can `X.X.X` as the version string.
+
+:::note
+Avoid using `X.X.X` unless necessary. Use the `X` placeholder only in the minor or patch position—for example, `1.X.X` or `1.2.X`—to maintain greater control over version updates.
+:::
+
+## Create a template version
+
+You can add new versions to your templates using the UI. The steps are the same for all resources.
```
kmadel, line 35: You can also add a new version via a GitOps approach by directly applying (kubectl, Argo CD, .) an update `VirtualClusterInstance` to the vCluster Platform cluster.

kmadel (changes requested):
I had a few recommendations.

platform/administer/templates/parameters.mdx
```diff
@@ -0,0 +1,431 @@
...
+<br /> 
+
+| Function      | Reason                          |
+|---------------|---------------------------------|
+| `env`         | Disabled to prevent env access  |
+| `envExpand`   | Disabled to avoid string injection from env vars |
+| `expandenv`   | Same as `envExpand`             |
+
+
+vCluster extends Helm’s function set with custom helpers for working with structured config formats.
```
lizardruss, line 289: nit: vCluster Platform (vCluster only might imply it's available with OSS)

joaniekube:
@deniseschannon PTAL

lizardruss (approved):
One optional nitpick, but looks great!

kmadel (changes requested):
Looks good overall, just a couple more suggestions.

joaniekube:
@deniseschannon PTAL

joaniekube:
@deniseschannon PTAL. Can this please be approved?

Restructured the flow so that it is clear what is for UI and what is for the manifests.

PR #924, opened 2025-07-28
Restructure docs to clean up navigation

deniseschannon:


Piotr1215 (approved):
@deniseschannon this looks great! There are no breaking changes in netlify.toml and links work we well. The changes should not affect partials generation either.

PR #931, opened 2025-07-31
Clarifies Rancher support and install instructions

cbron:
Not sure what that 2nd line was supposed to be.

### DIFF
--- a/vcluster/third-party-integrations/rancher/install-rancher-integration.mdx
+++ b/vcluster/third-party-integrations/rancher/install-rancher-integration.mdx
@@ -21,8 +21,8 @@ This guide explains how to install the vCluster Rancher Operator.
 Ensure you have the following prerequisites before installing the vCluster Rancher Operator:
 
 - Admin permissions for the local cluster. You must be able to add Helm repositories and install charts in the local cluster.
-- A running Rancher local cluster selected for installation
-- vCluster installed and at least one connected downstream cluster where virtual cluster are to be deployed.
+- A running Rancher local cluster selected for installation. Please note that the docker install is not support for production use cases.
+- At least one connected downstream cluster where virtual cluster are to be deployed.
 
 ## Install the vCluster Rancher Operator
 


vcluster/third-party-integrations/rancher/install-rancher-integration.mdx
```diff
@@ -21,8 +21,8 @@ This guide explains how to install the vCluster Rancher Operator.
 Ensure you have the following prerequisites before installing the vCluster Rancher Operator:
 
 - Admin permissions for the local cluster. You must be able to add Helm repositories and install charts in the local cluster.
-- A running Rancher local cluster selected for installation
-- vCluster installed and at least one connected downstream cluster where virtual cluster are to be deployed.
+- A running Rancher local cluster selected for installation. Please note that the docker install is not support for production use cases.
```
rmweir, line 24:
```suggestion
- A running Rancher local cluster selected for installation. Please note that the docker install is not supported for production use cases.
```
And do you mean rancher doesn't support it or we don't support it?

vcluster/third-party-integrations/rancher/install-rancher-integration.mdx
```diff
@@ -21,8 +21,8 @@ This guide explains how to install the vCluster Rancher Operator.
...
+- At least one connected downstream cluster where virtual cluster are to be deployed.
```
rmweir, line 25: Connected to what? Platform or Rancher? I think a user would think platform.
```suggestion
- At least one connected downstream cluster where virtual clusters are to be deployed.
```

rmweir (changes requested):

vcluster/third-party-integrations/rancher/install-rancher-integration.mdx
cbron: Both. We can't support it in prod if they don't.

vcluster/third-party-integrations/rancher/install-rancher-integration.mdx
cbron: I didn't actually write that part, but we could remove connected, it just refers to downstream cluster.

Piotr1215 (approved):

PR #957, opened 2025-08-05
add air-gapped docs for private nodes

hidalgopl:
# Content Description

## Internal Reference

Closes DOC-837 DOC-812 ENG-7410


deniseschannon (approved):
There are some changes within the air gap doc that I would make around how to upload your images to your node, but I'll revise that later.

deniseschannon (changes requested):
I re-read DOC-837 and you'll need to do more than just document the flag. 

We need to explain how to use it as part of managing upgrades that are upgrade. 

There should be some context in this page:

https://deploy-preview-957--next-vcluster-docs-site.netlify.app/docs/vcluster/next/deploy/worker-nodes/private-nodes/manage

PR #989, opened 2025-08-12
Standalone Docs

deniseschannon:
# Content Description

## Internal Reference

Closes DOC-


vcluster/deploy/control-plane/binary/readme.mdx
```diff
@@ -0,0 +1,278 @@
...
+
+<center>
+  <img
+    src="/docs/media/private-nodes/architecture.png"
+    width='500'
+    alt="vCluster Standalone Architecture"
+  />
+</center>
+
+You can connect and manage vCluster Standalone via vCluster Platform.
```
cbron, line 25: We should remove this, will cause problems. We should actually say to not add it to platform at this point.

cbron (changes requested):

PR #1119, opened 2025-09-05
fix: use the default bridge mode for the KubeVirt auto-nodes example

mfranczy:
# Content Description

Enhancement of KubeVirt VM example. For auto nodes users should use the bridge mode.
Otherwise all created guests will have the same IP address.

## Internal Reference

Closes DOC-

@netlify /docs

### DIFF
--- a/platform/administer/node-providers/kubevirt.mdx
+++ b/platform/administer/node-providers/kubevirt.mdx
@@ -116,16 +116,10 @@ spec:
                 - name: containerdisk
                   disk:
                     bus: virtio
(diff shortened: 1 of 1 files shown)


mfranczy:
> @mfranczy thanks for the PR. Did the PR template not trigger for you? It has a netlify setting that is needed to generate correct preview URL.

Sorry, my automation for creating PRs messed up the template, I will adjust it to respect the template.

FabianKramm (approved):

PR #1177, opened 2025-09-24
ENG-9061 | Change phrasing on private nodes upgrade description

janekbaraniewski:
Resolves ENG-9061


cbron (approved):
If this is still confusing we can create a section (or breakout a sentence) for this, describing what auto-upgrade is limited to, inside of the standalone docs.

PR #1217, opened 2025-10-13
feat: add inline algolia search to docs landing page

Piotr1215:
# Content Description
Adds an inline Algolia search box to the docs landing page, embedded directly in the intro card alongside the vCluster logo and description. Click or press <kbd>Ctrl/Cmd+K</kbd> to expand results in-page (no modal).

**Actual delta:** this PR only modifies `docs/index.mdx`. Version-aware search is provided by the SearchBar and SearchPage swizzles from #2013; the inline widget loads `/docs/search` in an iframe and inherits that behaviour.

### Search behavior
- Inline search box rendered inside the existing left-aligned intro-card on `/docs` (layout matches `main`)
- <kbd>Ctrl/Cmd+K</kbd> opens, <kbd>Esc</kbd> closes
- On the landing page the inline widget captures <kbd>Ctrl/Cmd+K</kbd> (capture phase, `stopPropagation`) so the DocSearch modal is suppressed; on all other pages the modal works normally
- Results are served by the version-aware `/docs/search` page (from #2013) embedded in an iframe — version boosting and stable-rank scoring apply automatically
- Debounced query updates; FOUC prevented via `isIframeReady` (iframe stays hidden until CSS injection completes)
- `BrowserOnly` wrapper for SSR safety

### Hardening (review pass)
- MutationObserver no longer leaks on every debounced query (was reattached on each iframe reload, never disconnected) — now tracked in a ref and disconnected on next load and unmount
- <kbd>Cmd+K</kbd> with CapsLock on now works (`e.key.toLowerCase()`)
- Cursor is no longer yanked back to the input on every iframe reload — focus is only restored if the iframe actually grabbed it
- Results panel uses `--ifm-background-surface-color` instead of hardcoded white (dark mode no longer breaks)
- A11y: `aria-label` on search input and close button, `aria-hidden` on decorative magnifier SVGs, semantic `<kbd>` for the keyboard hint

### Known limitation
Results are rendered inside an iframe. The version dropdowns from the search page are present but hidden by CSS injection, so users cannot change versions from within the widget — they need to navigate to the full search page for that. A future migration to `react-instantsearch` would eliminate the iframe and expose version controls inline.

### Conflict resolution with main
- Branch's own SearchBar swizzle superseded by #2013 during merge-from-main; version label rendering is now handled there
- Adopted main's repositioned intro copy (tenant clusters, AI clouds, hyperscaler-grade isolation)
- Dropped unused `Tabs`/`TabItem` imports

## Preview Link
https://deploy-preview-1217--vcluster-docs-site.netlify.app/docs

## Internal Reference
References DOC-972

@netlify /docs


guowenatk (dismissed):

Piotr1215:
Cross-linking with #2013 ("Improve DocSearch version-aware ranking") — they target the same UX problem but at different layers and overlap on three files:

- `src/theme/SearchPage/index.js` — #2013 has a broader swizzle (introduces `SEARCH_ALL_VERSIONS` / `SEARCH_CURRENT_VERSION` constants, defaults to "current"). My narrower "Stable" default would be superseded.
- `src/theme/SearchPage/styles.module.css` — both add styles.
- `docusaurus.config.js` — both add `contextualSearch: true`.

**Plan if #2013 lands first** (likely, since it's the foundational work):
- Drop my `SearchPage/index.js` and `styles.module.css` swizzles
- Drop the duplicate `contextualSearch: true` line
- Keep `docs/index.mdx` (the inline search box) — this is the unique contribution and it iframes into the `SearchPage` that #2013 produces

That collapses this PR to roughly: landing-page MDX + the BrowserOnly-wrapped `InlineSearch` component, sitting cleanly on top of #2013's infrastructure.

If this PR lands first instead, #2013 will need to reconcile its broader swizzle against my narrower, so #2013 first is preferable.

djwfyi (approved):
LGTM

PR #1574, opened 2026-01-13
Update backup.mdx - remove external DB limitation

matskiv:
# Content Description

The virtual clusters using an external database are supported by the current implementation of the snapshot/restore mechanism, so the documented limitation is out of date.
Current implementation is connecting to the backing store with an etcd client from a running vcluster pod where kine is running to expose the external database with etcd api.

## Preview Link 

https://deploy-preview-1574--vcluster-docs-site.netlify.app/docs/vcluster/next/manage/backup-restore/backup#limitations

## Internal Reference

@netlify /docs

> [!NOTE]
> Removes outdated limitation stating that virtual clusters with external databases cannot use vCluster's snapshot/restore mechanism. The current implementation supports external databases through kine's etcd API exposure.
>
> <sup>Generated by Claude for a21ee2c3375bccbdcecc940b9b8d2312660024a2</sup>

### DIFF
--- a/vcluster/manage/backup-restore/backup.mdx
+++ b/vcluster/manage/backup-restore/backup.mdx
@@ -230,8 +230,5 @@ When taking snapshots and restoring virtual clusters, there are limitations:
 **Sleeping virtual clusters**
 - Snapshots require a running vCluster control plane and do not work with sleeping virtual clusters.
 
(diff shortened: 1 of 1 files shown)


cbron (changes requested):
Not positive this actually works yet. We'll test first.

PR #1961, opened 2026-04-20
fix(llms-txt): strip react jsx comment markers from generated .md

Piotr1215:
# Content Description

React emits empty `<!-- -->` comment nodes around JSX expression boundaries (for example, around `<GlossaryTerm>` in the middle of a sentence). Those markers survive the HTML → Markdown pipeline in `@signalwire/docusaurus-plugin-llms-txt`, so the generated `.md` files end up with noise like:

```
put <!-- -->virtual cluster<!-- -->s to sleep
```

On this branch's local build, **131** `.md` files contained such artifacts. After this change, **0** do.

The fix is a tiny rehype plugin (`plugins/rehype-strip-comments.js`) that drops hast `comment` nodes. It is wired in via the `beforeDefaultRehypePlugins` slot on the llms-txt plugin, so it runs ahead of `rehype-remark` and only affects the `.md` output — the rendered site HTML is untouched.

## Why this matters beyond aesthetics

`afdocs`' `markdown-content-parity` check compares HTML text against the served `.md`. Its `normalize()` step converts `<!-- -->` into literal `!-- --` on the markdown side, while the HTML side strips comments entirely. That mismatch made any sentence containing a glossary term look like "missing content" to the parity checker, inflating the failure count.

Stripping the comment nodes at the rehype stage closes that class of false positives.

## Scope

- Only the three failure categories called out in DOC-1322; this PR addresses **category 1** (glossary `<!-- -->` noise).
- Categories 2 (custom React components without a markdown fallback) and 3 (admonition serialization drift) remain, and are further confounded by an afdocs / `node-html-parser` issue: Docusaurus v3.8.0 emits minified HTML without explicit `<head>`/`<body>` tags, which causes `node-html-parser` to fall back to `root.text` and not apply its block-tag separation or chrome-stripping logic. Those are better handled in follow-up work.
- No existing CI workflow runs parity checks — wiring one is out of scope here.

## Preview Link

- https://deploy-preview-1961--vcluster-docs-site.netlify.app/docs/vcluster/next/key-features/sleep (content unchanged; plugin config only)

## Internal Reference

Closes DOC-1322

@netlify /docs

## Regression test

Added a Playwright spec (`tests/specs/markdown-parity.spec.js`) that fetches three pages with dense glossary usage and asserts the served body contains neither `<!--` nor `-->`. Auto-picked up via `testDir: './specs'`. Run via the `run-e2e` label.

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -10,6 +10,7 @@ const __webpack_public_path__ = "/docs/";
 
 import resolveGlob from "resolve-glob";
 import remarkVersionTokens from "./plugins/remark-version-tokens.js";
+import rehypeStripComments from "./plugins/rehype-strip-comments.js";
 
 const newDocTemplate = `---
 title: Your Document Title
@@ -183,6 +184,11 @@ const config = {
           // surface these links to users in CLI contexts where relative paths
           // are not clickable or resolvable. See ENGAI-58.
           relativePaths: false,
+          // Strip React-emitted `<!-- -->` JSX expression markers before the
+          // HTML→Markdown conversion. Runs at the hast stage, ahead of
+          // rehype-remark, so comment nodes are gone before mdast is built.
+          // See DOC-1322.
+          beforeDefaultRehypePlugins: [rehypeStripComments],
         },
         siteTitle: 'vCluster Documentation',
         siteDescription: 'Documentation for vCluster (virtual Kubernetes clusters) and vCluster Platform (multi-cluster management)',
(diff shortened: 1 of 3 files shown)


Piotr1215:
Confirmed working with a test sample

```markdown
new
  3. In the drawer that appears from the right, click on the Agent configuration pane. Provide the values in the textarea under Extra Agent
  Values in the YAML format.

  compared to old one:
  3. In the drawer that appears from the right, click on the Agent <!-- -->configuration pane. Provide the values in the textarea under<!-- -->
  <!-- -->Extra Agent Values in the YAML format.
```

PR #2013, opened 2026-04-24
Improve DocSearch version-aware ranking

djwfyi:
# Content Description
This updates the docs search experience to be version-aware end to end. It adds explicit DocSearch metadata to doc pages, checks in a crawler configuration that can facet and rank by stable/current version, and swizzles the search UI so users can default to current stable docs while still filtering to other versions.

## Preview Link

https://deploy-preview-2013--vcluster-docs-site.netlify.app/docs/vcluster/

## Internal Reference
N/A

@netlify /docs


djwfyi:
## Summary of crawler config changes

Here’s the concrete delta from the current crawler config to this proposal.

1. Record metadata becomes version-aware
The current crawler relies almost entirely on the default Docusaurus DocSearch fields:
- `type`
- `lang`
- `language`
- `version`
- `docusaurus_tag`

The new approach adds explicit metadata emitted by the docs pages and captured by the crawler:
- `product`
- `version_label`
- `version_status`
- `stable_version`
- `is_stable`
- `is_latest`
- `page_category`

This makes it possible to rank and filter by “current stable” vs older versioned docs vs unreleased docs, instead of treating everything as generic Docusaurus content.

2. `weight.pageRank` is actually used
The current index settings already include:
- `desc(weight.pageRank)`

But the current extractor never sets `pageRank`, so that ranking signal is effectively unused.

The proposed crawler computes `pageRank` based on docs intent:
- highest for top-level product landing pages
- strong boost for stable/current docs
- lower for older versioned docs
- lowest for unreleased `next` docs

This keeps older versions searchable, but makes the current stable docs win more often.

3. Faceting is expanded for a tailored search UX
Current faceting:
- `type`
- `lang`
- `language`
- `version`
- `docusaurus_tag`

Proposed faceting adds:
- `searchable(product)`
- `searchable(version_label)`
- `filterOnly(version_status)`
- `filterOnly(stable_version)`
- `filterOnly(is_stable)`
- `filterOnly(is_latest)`
- `filterOnly(page_category)`

That enables a better UI flow:
- default to current stable
- optionally filter to a specific version
- optionally search across all versions

4. Search behavior is safer around versioned docs
The current setup is close to the stock Docusaurus template and depends on contextual search alone.

The new proposal keeps contextual search, but adds:
- explicit version metadata on every page
- stable-version boosting
- search-page version selectors for each docs product
- a default search bias toward current stable docs

5. Crawl scope is tighter
Current config includes:
- `startUrls` for `/docs/`, `/`, and `/docs/vcluster`
- `discoveryPatterns: ["https://www.vcluster.com/**"]`

The proposed config narrows discovery to docs-only URLs. That reduces noise and keeps the index focused on documentation content instead of broad site discovery.

6. Selectors stay mostly the same
This is not a wholesale rewrite of extraction. The heading/content selectors remain essentially the Docusaurus pattern:
- `lvl1: header h1`
- `lvl2` through `lvl5`
- `content: article p, article li, article td:last-child`

So the main difference is not extraction structure. The main difference is better metadata and better ranking logic.

In short: the current crawler is a mostly standard Docusaurus DocSearch config, while the proposed version turns it into a version-aware search index that can preferentially surface current stable docs without hiding older versions.

djwfyi:
## Deployment steps

Before merging, I’d do only the prep in Algolia, not the actual reindex yet.

Why: the crawler hits https://www.vcluster.com/docs, and the new PR is what adds the docsearch:* meta tags. If you reindex before that code is live, Algolia won’t see the new metadata, so you won’t get the new version-aware fields.

Recommended sequence:

1. Back up the current Algolia state.

   Save:
   - the current crawler config
   - current index settings for vcluster
   - any synonyms or rules you care about

   Reason: deleting an index removes records, settings, rules, and synonyms. Algolia says index deletion removes all of those.

   Sources: [Delete indices](https://www.algolia.com/doc/guides/sending-and-managing-data/manage-your-indices/how-to/deleting-multiple-indices), [Delete an index](https://www.algolia.com/doc/libraries/sdk/v1/methods/delete-index/)

2. Update the crawler configuration in Algolia.

   Paste in the new crawler config from the PR, but don’t run a full crawl yet.
   This is safe prep work:

   - new extraction logic is ready
   - new faceting/ranking settings are defined
   - crawl scope is narrowed to docs

3. Decide how to apply the new index settings.

   This is the important part: initialIndexSettings only apply on the first crawl of an index. Algolia explicitly says they won’t be re-applied later.
   Source: [initialIndexSettings](https://www.algolia.com/doc/tools/crawler/apis/configuration/initial-index-settings/)
   So you have two options:

   A. Option A: manually update the existing vcluster index settings in Algolia to match the new config
   B. Option B: delete and recreate the vcluster index so the crawler applies initialIndexSettings from scratch

   I’d recommend Option B if you want the crawler config to be the source of truth, but only if you’ve backed up rules/synonyms/settings first.

4. Merge the PR and let production deploy.
   Wait until the new docs code is live on https://www.vcluster.com/docs.
   That’s the moment when pages start emitting:

   - `docsearch:product`
   - `docsearch:version`
   - `docsearch:version_status`
   - `docsearch:is_stable`
   - etc.

5. Run a reindex immediately after deploy.

   Once production is live, trigger a crawl/reindex from the Crawler UI or API. Algolia documents a crawl/reindex endpoint for this.
   Source: [Start a crawl](https://www.algolia.com/doc/rest-api/crawler/start-reindex)

6. Verify extracted records in the index.
   Check that records now contain the new fields:

   - `product`
   - `version`
   - `version_label`
   - `version_status`
   - `stable_version`
   - `is_stable`
   - `is_latest`

   Also verify that the facets exist and are queryable.

7. Smoke-test production search.

   Verify:

   - default search favors current stable docs
   - version filters on /search behave correctly
   - old versions still appear when explicitly selected
   - unreleased/current preview docs don’t dominate results

## Summary 

So the short version is:

Before merge:

- back up current Algolia state
- update crawler config
- choose whether you’ll manually update settings or recreate the index

After merge/deploy:

- run the reindex
- verify records/facets
- smoke-test search

The one thing I would not do before merge is a full reindex, because production doesn’t have the new metadata yet.

djwfyi:
## Review feedback addressed

| Concern | Change |
|---|---|
| **IMPLEMENTATION.md diff markers** | Removed `+` prefixes from the source string in Step 5's JSON heredoc — it was invalid JSON that would have broken the crawler config on copy-paste |
| **Missing `exclusionPatterns`** | Added `exclusionPatterns: ["https://www.vcluster.com/docs/v0.19/**"]` to `docsearch.config.js` to match the runbook's JSON payload |
| **Hardcoded stable versions** | Added a comment in `docsearch.js` pointing to the runbook, and added an "After Each Stable Release" section to `IMPLEMENTATION.md` with a two-step update process |
| **Dead exports** | Removed `getDefaultSearchVersionName` and `isSearchVersionMatch` from `docsearch.js` |
| **Modal vs. search page filter difference** | Added comments in both `SearchBar/index.js` and `SearchPage/index.js` explaining the intentional distinction (soft optionalFilters boost in modal vs. hard disjunctive facet filters on search page) |
| **`algoliaHelper` recreation** | Wrapped `algoliaClient` and `algoliaHelper` in `useMemo` in `SearchPage/index.js`; moved the `result` listener into a `useEffect` with proper cleanup via `removeListener` |
| **Search pill color** | Changed from hardcoded `#ff6600`/`#f15c00` to `var(--ifm-color-primary)` / `var(--ifm-color-primary-dark)` — will inherit whatever color the brand reposition settles on |
| **Synonyms** | Added four bidirectional synonym pairs to `export-vcluster-settings.preview.json` covering the active terminology repositioning: virtual cluster ↔ tenant cluster, multi-tenancy ↔ tenant isolation, host cluster ↔ control plane cluster, neocloud ↔ AI cloud |

Piotr1215:
Cross-linking with #1217 — both PRs converge on version-aware search but at different layers.

**Overlap (would conflict on second merge):**
- `src/theme/SearchPage/index.js` — #1217 has a narrow swizzle (default to label containing `"Stable"`); this PR rewrites with `SEARCH_ALL_VERSIONS` / `SEARCH_CURRENT_VERSION` constants and "current" defaults. Different solutions to the same problem.
- `src/theme/SearchPage/styles.module.css` — both add styles.
- `docusaurus.config.js` — both add `contextualSearch: true`; this PR also sets `searchPagePath` and replaces `externalUrlRegex`.

**Complementary (no overlap):**
- This PR owns the search infrastructure: crawler config, DocSearch metadata on doc pages, custom `SearchBar`.
- #1217 owns the landing-page UX: an inline `InlineSearch` component on `/docs` that iframes `/docs/search` (i.e., the page rendered by this PR's `SearchPage`).

**Proposed merge order:** this PR first. After it lands, #1217 rebases and drops:
- its `SearchPage` swizzle and `styles.module.css` (your version supersedes)
- the duplicate `contextualSearch: true` in `docusaurus.config.js`

That leaves #1217 as a pure landing-page UX change on top of your search infrastructure.

algolia/IMPLEMENTATION.md
```diff
@@ -0,0 +1,294 @@
...
+The docs implementation adds:
+
+- version-aware search metadata to doc pages via `meta` tags
+- a better search page UX with explicit version filters
+- a more visible DocSearch modal CTA to open the full search page
+- default search bias toward current stable docs
+
+Relevant files:
+
+- [docusaurus.config.js](/Users/daryl/git/vcluster/vcluster-docs/docusaurus.config.js)
```
Piotr1215: Maybe make these repo-relative — same pattern repeats throughout the file.
```suggestion
- [docusaurus.config.js](../docusaurus.config.js)
```

src/theme/SearchPage/index.js
```diff
@@ -0,0 +1,533 @@
...
+
+function SearchVersionSelectList({docsSearchVersionsHelpers}) {
+  const versionedPluginEntries = Object.entries(
+    docsSearchVersionsHelpers.allDocsData,
+  ).filter(([, docsData]) => docsData.versions.length > 1);
+
+  return (
+    <div className={styles.searchVersionGrid}>
+      {versionedPluginEntries.map(([pluginId, docsData]) => {
+        const displayNames = {vcluster: 'vCluster', platform: 'Platform'};
```
Piotr1215: Maybe add a displayName field to DOCSEARCH_PRODUCTS in src/config/docsearch.js so SearchPage and SearchBar share one source of truth.
```suggestion
        const displayName = getDocsearchProduct(pluginId)?.displayName ?? pluginId;
```

src/theme/SearchPage/index.js
```diff
@@ -0,0 +1,533 @@
...
+  const makeSearch = useEvent((page = 0) => {
+    algoliaHelper.clearRefinements();
+
+    if (contextualSearch) {
+      algoliaHelper.addDisjunctiveFacetRefinement('docusaurus_tag', 'default');
+      algoliaHelper.addDisjunctiveFacetRefinement('language', currentLocale);
+
+      Object.entries(docsSearchVersionsHelpers.searchVersions).forEach(
+        ([pluginId, searchVersion]) => {
+          if (searchVersion === SEARCH_ALL_VERSIONS) {
```
Piotr1215, line 313: This uses Docusaurus 'isLast', while SearchBar/optionalFilters and the docsearch:stable_version meta tag use product.stableVersion from src/config/docsearch.js. They align after release-day, but during rc-1 they can drift. Pull from getDocsearchProduct(pluginId).stableVersion for one source of truth.
```suggestion
          const latestStableVersion =
            getDocsearchProduct(pluginId)?.stableVersion ??
            pluginData.versions.find((version) => version.isLast)?.name ??
            pluginData.versions[0]?.name;
```

src/theme/SearchPage/index.js
```diff
@@ -0,0 +1,533 @@
...
+    }
+
+    return () => true;
+  }, [loaderRef]);
+
+  useEffect(() => {
+    searchResultStateDispatcher({type: 'reset'});
+    if (searchQuery) {
+      searchResultStateDispatcher({type: 'loading'});
+      setTimeout(() => {
```
Piotr1215: Each query change schedules a new setTimeout but doesn't clear the previous one — rapid typing fires overlapping Algolia requests. The reducer's stale-result guard catches this, but adding clearTimeout skips the wasted calls.
```suggestion
      const timer = setTimeout(() => {
        makeSearch();
      }, 300);
      return () => clearTimeout(timer);
```

Piotr1215 (commented):
lgtm, left a few comments

Piotr1215 (approved):

PR #2073, opened 2026-05-08
Update supported navBarButtons positions

vaidaslamanauskas:
From version 4.9, the platform sidebar (_side navigation_) supports only two `navBarButtons` positions, Top & Bottom.

@netlify /docs


djwfyi (approved):
LGTM

djwfyi:
Before I merge, a question:
Do users need to update their configs to say Top or Bottom, or will the old configs continue to work in 4.9?

vaidaslamanauskas:
> Before I merge, a question: Do users need to update their configs to say Top or Bottom, or will the old configs continue to work in 4.9?

No extra input from the user will be needed. Existing configs will work.

PR #2102, opened 2026-05-12
fix(platform): create dockerconfigjson secret with proper init value

seanschneeweiss:
# Content Description

The code block in the secret sync docs used the deprecated `kubernetes.io/dockercfg` type. Today, `kubernetes.io/dockerconfigjson`is more common, but requires the `.data.\.dockerconfigjson` field. This
  caused any user who copied the template to create an invalid secret that Kubernetes would reject.

Changes                                                                                                                                    
- Corrects secret type to kubernetes.io/dockerconfigjson
- Adds data[.dockerconfigjson]: e30K (base64 of {}) as a valid initialization value required by the type   

## Internal Reference

Part of ENGPLAT-124

@netlify /docs

### DIFF
--- a/platform/use-platform/secrets/secret-sync.mdx
+++ b/platform/use-platform/secrets/secret-sync.mdx
@@ -7,8 +7,8 @@ import FeatureTable from '@site/src/components/FeatureTable';
 <FeatureTable names="secrets" />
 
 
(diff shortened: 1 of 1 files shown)


djwfyi (approved):
LGTM for the purposes of this PR

PR #2188, opened 2026-05-27
feat(receiver): alert slack on docs-sync receiver failure

Piotr1215:
# Content Description
Adds an `if: failure()` Slack notification step to `handle-source-release.yml` so silent receiver failures stop being silent. Reuses the org's `ci-test-notify` composite action, not an inline `slackapi/slack-github-action`.

## Summary
- Append a `Notify Slack on receiver failure` step at the end of the `sync` job.
- Fires only when the run fails AND the classifier did not signal `skip=true` (alpha/beta versions are green-by-design no-ops and must never page anyone).
- Pinned by SHA `85d7023c5749421d369f59430c7849f2d00ad694 # ci-test-notify/v1` to match every other caller in the org.
- Webhook from `secrets.DOCS_SYNC_SLACK_WEBHOOK_URL` (TODO: provision, see below).

## Why now
The handle-source-release receiver runs unattended on every loft-sh release dispatch. Until this PR, a failed run produced no signal:
- Release went green (sender's dispatch step is `continue-on-error`).
- Receiver failed in the vcluster-docs Actions tab.
- Docs drift went unnoticed until someone looked.

Today's backfill of v0.34.1-rc.1 + v4.9.1-rc.1 surfaced this: zero `repository_dispatch` runs landed between 2026-05-08 and 2026-05-27 (token rotation gap nobody noticed), and the platform v4.9.1-rc.1 backfill failed twice at the api-pin bump step before #2184 shipped the goprivate-narrowing fix. With this step in place, both classes of failure (sender-token gaps surface indirectly via "expected receiver run never landed", and any receiver-step failure) would page the chosen channel directly.

## Key Changes
- `.github/workflows/handle-source-release.yml`: appended one composite-action step plus a leading comment explaining the skip-guard and the optional-webhook semantics.

## Dependencies
- Provisioning of `secrets.DOCS_SYNC_SLACK_WEBHOOK_URL`. Until then, `ci-test-notify` no-ops with a `::warning::` (per its `Check webhook URL` step), so this PR is safe to merge without the secret.

## Scope notes
- **Receiver-side only.** Failures in the sender (e.g. the HTTP 403 token-rotation gap that just bit us) do not produce a receiver run, so they bypass this alert. A sender-side alert (in `loft-sh/vcluster/.github/workflows/release.yaml` "Notify vcluster-docs of vcluster release" step and the loft-enterprise equivalent) is a separate concern, flagged for follow-up, not blocking this PR.
- **Prior plan closed out.** The PrometheusRule + `github-actions-exporter` route from DEVOPS-892 was deferred because the metric source isn't scraped today. This alert ships the lightweight alternative; the `alerts/` folder in `eng-infra-config` stays empty as already verified.

## TODO
- [ ] Piotr: provision the Slack incoming webhook in the chosen channel.
- [ ] Piotr: add the URL as `DOCS_SYNC_SLACK_WEBHOOK_URL` repo secret on `loft-sh/vcluster-docs`.
- [ ] Validate post-provisioning by triggering one deliberate failed `workflow_dispatch` run.
- [ ] Comment closeout on DEVOPS-892 noting this alert replaces the deferred PrometheusRule leg.

### Open question for review
**Channel target.** Brief listed `#loft-rocks-alerts` (existing receiver) vs new `#docs-alerts`. No prior discussion found in Slack search. Default to whichever fits org SOPs; webhook secret is wired through the action so changing target later is a webhook swap, not a code change.

## Preview Link
N/A, workflow-only change, no docs touched.

## Internal Reference
Closes DEVOPS-904

References #2184 (the goprivate-narrowing fix that unblocked the platform path while this alert was being scoped).

@netlify /docs


djwfyi (approved):
LGTM

I'm in favor of #docs-alerts for the channel things go to.

Piotr1215:
I would prefer not to create new notifications channels, let's use `[#ci-tests-alerts](https://loft-labs-inc.slack.com/archives/C08JT9DPFNF)`

PR #2244, opened 2026-06-05
docs: add Gateway API guidance

djwfyi:
# Content Description
Add Gateway API documentation for vCluster endpoint access, tenant route syncing, shared Gateway imports, and ingress-nginx migration planning.

Update related sleep mode, ingress access, Platform domain, and sync docs to link the new Gateway API guidance and clarify controller support requirements.

## Preview Link 
- https://deploy-preview-2244--vcluster-docs-site.netlify.app/docs/vcluster/next/key-features/gateway-api
- https://deploy-preview-2244--vcluster-docs-site.netlify.app/docs/vcluster/next/manage/ingress-nginx-to-gateway-api
- https://deploy-preview-2244--vcluster-docs-site.netlify.app/docs/vcluster/next/configure/vcluster-yaml/sync/to-host/networking/gateway-api
- https://deploy-preview-2244--vcluster-docs-site.netlify.app/docs/vcluster/next/configure/vcluster-yaml/sync/from-host/gateways

## Internal Reference
Closes DOC-1487

Validation:
- vale platform/configure/installation-options/domain.mdx platform/use-platform/virtual-clusters/key-features/sleep-mode.mdx vcluster/_fragments/sync-from-host-resources.mdx vcluster/_fragments/sync-to-host-resources.mdx vcluster/_partials/sleep/activity-detection.mdx vcluster/configure/vcluster-yaml/control-plane/deployment/ingress.mdx vcluster/configure/vcluster-yaml/sleep.mdx vcluster/key-features/ingress-access.mdx vcluster/key-features/sleep.mdx vcluster/manage/accessing-vcluster.mdx vcluster/configure/vcluster-yaml/sync/from-host/gateways.mdx vcluster/configure/vcluster-yaml/sync/to-host/networking/gateway-api.mdx vcluster/key-features/gateway-api.mdx vcluster/manage/ingress-nginx-to-gateway-api.mdx
- npm run validate-mdx-links
- npm run build

AI review: mention `@claude` in a comment to request a review or changes. See [CONTRIBUTING.md](https://github.com/loft-sh/vcluster-docs/blob/main/CONTRIBUTING.md#ai-assisted-pr-review) for available commands.

> FORK LIMITATION: `@claude` does not work on pull requests opened from forks. GitHub Actions cannot access the required secrets for fork-originated PRs. To use AI review, push your branch directly to this repository.

@netlify /docs

PR #2321, opened 2026-06-22
fix(generated-files): derive config required flag from schema

djwfyi:
# Content Description
The platform and vCluster config reference pages labeled almost every field "required" (in both the body badge and the right-hand table of contents) because the generator defaulted every field to required and always emitted the literal "required" text, relying on CSS to hide it. This fixes the generator to read each field's actual required value from the schema and to emit the badge text only for genuinely required fields, then regenerates the reference partials and backports them to platform 4.10.0 and vCluster 0.35.0.

## What changed
- **Generator (`hack/`)**: required flag now derived from the schema `required` array (platform: via the `omitempty` convention; vCluster: from `vcluster.schema.json` required arrays). Badge text is conditional so the TOC reflects it. Nullable-field description fallback added. Gateway API sync prose moved into `pathExtras` so it survives regeneration.
- **Platform**: `status_reference.mdx` 114 → 29 required; backported to 4.10.0.
- **vCluster**: 0 → 384 required (genuinely required fields: patch path/apiVersion/kind, node pool name/quantity, autoNodes provider, etc.); backported to 0.35.0.

Note: This PR also backports the changes to both 0.35 and 4.10 versioned docs. Since the change applies to both sets of docs, and the backport CI doesn't handle multiple component backports.

## Known issue (tracked upstream)
Regenerating the vCluster partials reverts "tenant cluster" → "virtual cluster" in two field descriptions (snapshots, deletion) because the upstream schema comments still use the old term. This will recur on every regen until upstream is fixed.

## Preview Link

https://deploy-preview-2321--vcluster-docs-site.netlify.app/docs/platform/next/configure/platform-configs/overview/#auth-github

This is the page called out in the issue, but it's worth poking around to make sure the fix applies correctly throughout. Also updates the vCluster generated docs.

## Internal Reference
Closes DOC-1526

Upstream follow-up for the missing `omitempty`/`+optional` tag hygiene: ENGCP-970

@netlify /docs

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/hack/vcluster/partials/main.go
+++ b/hack/vcluster/partials/main.go
@@ -176,7 +176,6 @@ func main() {
 			"(positional args still accepted for back-compat)")
 	}
 
-	util.DefaultRequire = false
 	jsonSchemaPath := filepath.Join(versionDir, "vcluster.schema.json")
 	defaultValues := filepath.Join(versionDir, "default_values.yaml")
 	values, err := os.ReadFile(defaultValues)
--- a/hack/vcluster/partials/main.go
+++ b/hack/vcluster/partials/main.go
@@ -46,6 +46,11 @@ var pathExtras = map[string]Extras{
 			"`joinNode.enabled: false` and use separate worker nodes for tenant workloads.\n" +
 			":::\n\n",
 	},
+	"sync/toHost/gatewayApi": {
+		Before: "\nSetting `enabled: true` turns on Gateway and HTTPRoute sync, imports control plane " +
+			"cluster GatewayClasses so tenant Gateways can resolve them, and serves tenant ReferenceGrants " +
+			"for validation; TLSRoutes and BackendTLSPolicies must be enabled individually.\n\n",
+	},
 }
 
 // we only generate paths we actually need
--- a/hack/vcluster/partials/main.go
+++ b/hack/vcluster/partials/main.go
@@ -49,7 +49,9 @@ var pathExtras = map[string]Extras{
 	"sync/toHost/gatewayApi": {
 		Before: "\nSetting `enabled: true` turns on Gateway and HTTPRoute sync, imports control plane " +
 			"cluster GatewayClasses so tenant Gateways can resolve them, and serves tenant ReferenceGrants " +
-			"for validation; TLSRoutes and BackendTLSPolicies must be enabled individually.\n\n",
+			"for validation; TLSRoutes and BackendTLSPolicies must be enabled individually.\n\n" +
+			"In auto mode grants follow route sync and are validated within the tenant cluster; " +
+			"they sync to the control plane cluster only when namespace sync is also enabled.\n\n",
 	},
 }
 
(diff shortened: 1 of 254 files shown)


djwfyi:
Note: I identified a separate issue while working on this that prevented fully expanding nested keys. There's a separate PR to fix that at #2323.

hack/vcluster/partials/main.go
```diff
@@ -46,6 +46,11 @@ var pathExtras = map[string]Extras{
 			"`joinNode.enabled: false` and use separate worker nodes for tenant workloads.\n" +
 			":::\n\n",
 	},
+	"sync/toHost/gatewayApi": {
+		Before: "\nSetting `enabled: true` turns on Gateway and HTTPRoute sync, imports control plane " +
```
Piotr1215, line 50: This migrates the Gateway API prose into pathExtras so it survives regen, but only one of the two sentences from #2244 made it. The referenceGrants `auto` mode sentence is now dropped everywhere and won't come back on future regens. Restoring it here (or as an After on the referenceGrants path, which is its original spot) keeps it.
```suggestion
		Before: "\nSetting `enabled: true` turns on Gateway and HTTPRoute sync, imports control plane " +
			"cluster GatewayClasses so tenant Gateways can resolve them, and serves tenant ReferenceGrants " +
			"for validation; TLSRoutes and BackendTLSPolicies must be enabled individually.\n\n" +
			"In auto mode grants follow route sync and are validated within the tenant cluster; " +
			"they sync to the control plane cluster only when namespace sync is also enabled.\n\n",
```

Piotr1215 (commented):
Complements the automated panel review above (which covered correctness, the format-verb alignment, dead-code removal, and the three PR-metadata notes). Two content-fidelity findings the panel did not surface, both non-blocking:

1. Dropped prose (inline below). Migrating the Gateway API guidance into `pathExtras` carried over only one of the two sentences added in #2244. The referenceGrants `auto` mode sentence is gone from the regenerated output and won't survive future regens.

2. Scope FYI: the platform `allowedNodeTypes` field (projects reference) also changed shape in this regen, from `object[]` with a nested `name` child to `array|null` with a fuller description. That's the regen picking up a newer upstream type, not the required-flag change, and the new rendering looks correct. Worth a line in the PR body so the shape change isn't mistaken for a generator bug.

(The `tenant cluster` to `virtual cluster` revert is already well covered by the panel's third note, so not repeating it.)

hack/vcluster/partials/main.go
djwfyi: Good catch. Restored the referenceGrants auto-mode sentence to the gatewayApi `pathExtras` `Before` block (your wording verbatim) so it survives regen, then regenerated the unversioned and 0.35.0 leaf partials. Both sentences now render on the gateway-api reference page. Pushed in 6596431b0.

DadaGore:
I'm not fully across how this is meant to work ideally, but as long as the "required" bit is removed from the config files, it looks good to me.

PR #2322, opened 2026-06-22
feat(vcluster-yaml): prototype kubectl-doc interactive schema explorer

Piotr1215:
# Content Description

Adds an interactive, YAML-shaped schema explorer for `vcluster.yaml` (kubectl-doc `KubeSchemaDoc`, sttts/kubectl-doc, Apache-2.0) and surfaces it as an additional view on the `sync/to-host` config reference page, beside the existing accordions. Includes the section-to-payload generator and a full evaluation. DOC-1512.

## Summary

The big config pages are where the accordion reference strains (`controlPlane` is 5,425 lines / 357 fields). The explorer renders the schema as the YAML you actually write, with a details panel, folding, and an in-tree filter. Following review from the docs team, it ships as an augmentation: the generated accordions stay the default, indexed reference, and an "Explorer (beta)" tab sits next to them. `kubectl doc` cannot ingest the vcluster JSON Schema directly (it takes a Kubernetes CRD or cluster OpenAPI), so a generator converts a schema section to a synthetic CRD, runs the real kubectl-doc renderer, and strips the k8s frame back to the section root.

## Key Changes

- `vcluster/configure/vcluster-yaml/sync/to-host/README.mdx`: `## Config reference` is now a `<Tabs>` with Reference (accordions, default) and Explorer (beta). First real-page rollout.
- `hack/vcluster/schema-explorer/generate-payloads.mjs`: section-to-payload automation. Drives kubectl-doc (faithful `lines[]` tokens, no hand-rolled tokenizer), strips `apiVersion`/`kind`/`metadata`/`spec` and rewrites paths to real dotted paths (`sync.toHost.*`).
- `src/components/kubectl-doc/`: vendored `KubeSchemaDoc` (v0.2.9, Apache-2.0), SSR-safe wrapper, `MountWhenVisible` (lazy-mount on tab open), navbar positioning override, and the generated `sync/to-host` payload under `vcluster/_partials/config/`.
- `hack/vcluster/schema-explorer/EVALUATION.md`: findings for all acceptance criteria, source-of-truth/versioning, go/no-go, risks.

## How it stays current

The source of truth is `chart/values.schema.json` in `loft-sh/vcluster` (generated upstream from the Go config). The `configsrc/vcluster/<version>/` copies here are per-release snapshots, not the source. Regeneration mechanism (decided): drive the real kubectl-doc renderer per section, so payloads are faithful by construction. The next step wires `generate-payloads.mjs` into the existing release automation (`handle-source-release.yml` to `run-generator.sh`, `vcluster-released` case) with per-version output and a `git diff --exit-code` drift gate, exactly like the accordion partials. That is the one change that touches a release workflow, so it lands as a separate, clearly-scoped follow-up.

## Verification

Verified on the `next` docs build (`/docs/vcluster/next/configure/vcluster-yaml/sync/to-host`): both tabs render, Reference carries all 390 accordion fields (server-rendered, search-indexed), Explorer lazy-mounts on view, field paths and deep-link hashes are correct (`sync.toHost.configMaps`), zero console errors.

## Risks / caveats

- Client-only rendering means no SSR content for the explorer. The site uses Algolia DocSearch (index `vcluster`), which crawls rendered HTML, so explorer-only fields are not in site search or Google. Mitigated by keeping the accordions as the default, indexed tab.
- Vendors an 87 KB generated runtime (`kubectl-doc-runtime.js`), pinned and Apache-2.0, provenance recorded. Lazy-mounted so only pages with the tab load it, and only when opened.
- `oneOf`/`anyOf`/`allOf` collapse in the CRD adapter is lossy; the pilot section (`sync.toHost`) is union-free. Audit before extending to sections with unions.

## Dependencies

None.

## TODO

- [ ] Wire `generate-payloads.mjs` into `run-generator.sh` (`vcluster-released`) with a `check-generated` drift gate and the `go install kubectl-doc@v0.2.9` step.
- [ ] Roll the Explorer tab to more overview pages (`control-plane`, `sync/from-host`) once the pilot is approved.
- [ ] Design team: polish the detail-card styling (positioning stopgap is in place).

## Preview Link

- Explorer tab on a real page: https://deploy-preview-2322--vcluster-docs-site.netlify.app/docs/vcluster/next/configure/vcluster-yaml/sync/to-host
- Standalone sandbox: https://deploy-preview-2322--vcluster-docs-site.netlify.app/docs/kubectl-doc-prototype

## Internal Reference
Closes DOC-1512

@netlify /docs


djwfyi (changes requested):
I'm pretty excited about this one. A few things I note:

1. The explorer payload is loaded with the page, so it always gets pulled. It might be better to have it load if someone clicks on the tab instead so not everyone has to download the extra 907 KB. The runtime lazy loads, but the payload does not. I'm not sure how important this is to fix, but I had Codex come up with a suggestion I'll post as a separate comment. It makes implementing this broadly a touch more involved. Which may make it easier to just keep things as they are now.

2. The explorer currently renders with `configMaps` looking like a top level key. It should be nested under `sync.toHost`. The pop up has the full key path, but it would be nice to have the YAML structure look correct on the render, too.

3. `docs/kubectl-doc-prototype` is getting added to the sitemap. We should exclude it in `docusaurus.config.js` sitemap section's `ignorePatterns` block. Or we could just drop the page since we have a live example of it in use now.
  
  ```
   ignorePatterns: [
     '/kubectl-doc-prototype',
     '/kubectl-doc-prototype/**`,
     ...
  ```

djwfyi:
One suggestion for the lazy-load piece: I would avoid making the page statically import the generated explorer JSON, because that still puts the full payload in the route chunk before the user opens the Explorer tab. I would also avoid a central dynamic-import registry, since that gets fussy as this rolls out to more sections.

A more scalable Docusaurus shape is to have the generated payloads emitted as static assets and let the explorer receive a URL:

```mdx
import LazySchemaExplorer from '@site/src/components/kubectl-doc/LazySchemaExplorer';

<MountWhenVisible>
  <LazySchemaExplorer src="/docs/schema-explorer/vcluster/sync/toHost.explorer.json" />
</MountWhenVisible>
```

Then `LazySchemaExplorer` can just `fetch(src)` after `MountWhenVisible` renders it, and pass the parsed JSON to `SchemaExplorer`. That keeps both the 87 KB runtime and the much larger generated payload off the initial page load, while adding a new page only requires generating another JSON file and passing a different URL. No central switch statement, no top-level JSON imports in MDX, and no route chunk bloat as this expands.

Piotr1215:
Thanks for the review, all three points are addressed in fcbc6e5f.

1. Payload no longer downloads with the page. The generated payload moved out of the MDX route chunk to a static asset (`static/schema-explorer/vcluster/sync/toHost.explorer.json`), and a new `LazySchemaExplorer` fetches it at runtime. Paired with `MountWhenVisible`, neither the 87 KB runtime nor the ~800 KB payload loads until a reader opens the Explorer tab. This follows your suggested shape exactly: static asset plus a `src` URL, no top-level JSON import in MDX, and no central dynamic-import registry. Adding a section is now just "generate another JSON under `static/` and point a new `<LazySchemaExplorer src="..."/>` at it."

2. The explorer now renders the real YAML structure. `make-crd.mjs` nests the section under its true ancestor path (`spec.sync.toHost.*`) and carries each ancestor's schema description, so the tree renders as `sync:` -> `toHost:` -> `<fields>` instead of hoisting `configMaps` to the root. `generate-payloads.mjs` then just drops the CRD frame and the leading `spec.` segment. Verified on the regenerated payload: 0 fields leak at the top level, all 396 nest under `sync.toHost`, and the intermediate `sync`/`toHost` keys stay clickable with their real descriptions.

3. `kubectl-doc-prototype` is excluded from the sitemap. Added `/docs/kubectl-doc-prototype` and `/docs/kubectl-doc-prototype/**` to the `ignorePatterns` block. They need the `/docs` prefix because `ignorePatterns` matches `route.path`, which includes the site baseUrl. Kept the sandbox page for now since it is a useful isolated harness; it just no longer competes in search.

Re-requesting review.

djwfyi (approved):
All three points from the previous round are resolved and I verified each one rather than just reading the diff.

1. Lazy loading: confirmed the mechanism end-to-end. Docusaurus renders inactive `TabItem` panels with the `hidden` attribute, so `MountWhenVisible`'s `IntersectionObserver` never fires until the Explorer tab is opened, and `LazySchemaExplorer` only fetches the static JSON at that point. No stale imports of the old `_partials/config/sync/toHost.explorer.json` path remain anywhere in the repo.

2. Nesting: installed `kubectl-doc@v0.2.9` and regenerated the payload from scratch. Output is byte-for-byte identical to the committed file. `sync` is the only top-level key, `configMaps` no longer leaks to the root, and all 397 fields resolve to real paths like `sync.toHost.configMaps.enabled`.

3. Sitemap: `/docs/kubectl-doc-prototype` and `/docs/kubectl-doc-prototype/**` are correctly excluded via `ignorePatterns`.

No new issues in this commit. Nice work tightening this up.

PR #2340, opened 2026-06-29
feat(ci): open a docs follow-up when loft-sh/plans changes affect sidebar labels

Piotr1215:
# Content Description

Adds a weekly CI workflow that opens (or updates) a docs follow-up issue whenever `loft-sh/plans` changes, prompting a review of the left-nav Enterprise/Free sidebar labels and the `ProAdmonition` partial.

## Summary

The Feature Table already syncs plan tiers into `src/data/products.yaml` automatically, but the sidebar labels (`sidebar_class_name: pro|free`, rendered as ENTERPRISE/FREE badges by `src/css/sidebar.scss`) and the `ProAdmonition` partial are maintained by hand. When `loft-sh/plans` moves a feature between Free and Enterprise, nothing prompts a review of those surfaces, so the docs drift. DOC-1513 (custom resources wrongly marked Enterprise) is the exact failure this leaves uncaught.

This opens a review issue rather than an auto-PR. Unlike the mechanical EOS/EOL date drift (which keeps its auto-PR path in `check-eos-eol-labels`), the plan-tier to page-label mapping needs human judgement: a page can document several features, a `docs_url` can point at a section anchor, and a `ProAdmonition` is placed editorially. The issue carries a best-effort suspect list plus the full label inventory, so the reviewer gets a concrete starting point instead of an empty checklist.

## Key Changes

- `scripts/check-plans-sidebar-labels.js`: detects recent `loft-sh/plans` commits, cross-checks each feature tier in `products.yaml` against the label on its docs page (flags a Free feature whose page still carries `pro`/`ProAdmonition`, and an Enterprise-only feature whose page carries `free`), builds the full label inventory, and opens or updates a single deduped tracking issue.
- `.github/workflows/check-plans-sidebar-labels.yml`: weekly schedule + `workflow_dispatch` + `repository_dispatch` (`plans-updated`, so `loft-sh/plans` can trigger on push). Single script invocation, pinned action SHAs, `persist-credentials: false`.
- `tests/check-plans-sidebar-labels.test.js`: 24 unit tests (Node built-in runner, no new dependencies) covering frontmatter parsing, the cross-check in both directions, the inventory, commit parsing, issue-body rendering, and the open-or-update logic with an injected `gh` exec.
- `package.json`: a `check-plans-sidebar-labels` script for local dry-runs.

Validated locally against real repo data: the cross-check surfaces a genuine suspect (`external-database-kerberos` is Free in plans but its page carries `sidebar_class_name: pro` + a `ProAdmonition`).

## Preview Link

No documentation pages changed (CI tooling, script, and tests only), so there is no deploy preview for this PR.

## Dependencies

None. Reuses the existing `GH_ACCESS_TOKEN` secret, the same token the Feature Table sync uses to read the private `loft-sh/plans` repo.

## TODO

- [ ] Optional follow-up: add a `repository_dispatch` sender in `loft-sh/plans` to trigger the review on push. The workflow already runs weekly without it.

## Internal Reference

Closes DOC-1523
References DOC-1513

@netlify /docs

### DIFF
--- a/.github/workflows/check-plans-sidebar-labels.yml
+++ b/.github/workflows/check-plans-sidebar-labels.yml
@@ -16,6 +16,8 @@ on:
   schedule:
     # Weekly on Mondays at 09:27 UTC. Off-hour offset to avoid the top-of-hour
     # stampede with the feature-table sync (09:00) and EOS/EOL sync (09:13).
+    # The script looks back ~2 weeks (DEFAULT_SINCE_DAYS), so a skipped or
+    # delayed weekly run does not drop plans commits that landed in the gap.
     - cron: '27 9 * * 1'
   workflow_dispatch:
   repository_dispatch:
@@ -47,8 +49,20 @@ jobs:
         run: npm ci
 
       - name: Check plans sidebar labels
+        id: check
         # GH_ACCESS_TOKEN reads the private loft-sh/plans repo and files the
         # review issue in this repo (mirrors sync-upstream-features.yml).
         env:
           GH_TOKEN: ${{ secrets.GH_ACCESS_TOKEN }}
         run: node scripts/check-plans-sidebar-labels.js --open-issue
+
+      - name: Fail if the plans fetch failed
+        # A failed read of loft-sh/plans reports drift=false, which on a
+        # scheduled run is indistinguishable from a genuinely quiet week. The
+        # script emits plans_fetch_failed=true on that path; turn the job red so
+        # a revoked GH_ACCESS_TOKEN or lost repo access is visible instead of
+        # silently skipping the sidebar-label review. See DOC-1523.
+        if: steps.check.outputs.plans_fetch_failed == 'true'
+        run: |
+          echo "::error::Could not read loft-sh/plans commits; the sidebar-label review was skipped. Check that GH_ACCESS_TOKEN still has read access to loft-sh/plans."
+          exit 1
--- a/.github/workflows/check-plans-sidebar-labels.yml
+++ b/.github/workflows/check-plans-sidebar-labels.yml
@@ -1,0 +1,54 @@
+name: Check plans sidebar labels
+
+# When loft-sh/plans changes, the left-nav Enterprise/Free sidebar labels and
+# the ProAdmonition partial can drift: unlike the Feature Table they are hand
+# maintained, with no machine mapping from a plan tier to the exact pages that
+# need a label flipped. So this opens (or updates) a single review issue
+# carrying a best-effort suspect list and the full label inventory for a human
+# to act on. Mechanical drift gets an auto-PR (check-eos-eol-labels.yml);
+# judgement drift gets a review issue. See DOC-1523.
+#
+# All logic lives in scripts/check-plans-sidebar-labels.js (unit tested); this
+# workflow is a single invocation. The script treats workflow_dispatch and
+# repository_dispatch as a forced run via GITHUB_EVENT_NAME.
+
+on:
+  schedule:
+    # Weekly on Mondays at 09:27 UTC. Off-hour offset to avoid the top-of-hour
+    # stampede with the feature-table sync (09:00) and EOS/EOL sync (09:13).
+    - cron: '27 9 * * 1'
+  workflow_dispatch:
+  repository_dispatch:
+    # loft-sh/plans can POST a dispatch of this type to trigger a review on push.
+    types: [plans-updated]
+
+permissions:
+  contents: read
+  issues: write
+
+concurrency:
+  group: check-plans-sidebar-labels
+  cancel-in-progress: false
+
+jobs:
+  review:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
+        with:
+          persist-credentials: false
+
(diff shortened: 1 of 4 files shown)


djwfyi (commented):
A few suggested fixes:

`scripts/check-plans-sidebar-labels.js` -- silent failure in fetchPlansCommits

If the gh api call throws (e.g. token lacks read access to `loft-sh/plans`), `commits = []` and on a non-forced run the script outputs `drift=false` -- indistinguishable from a genuine no-activity week. Track the failure separately:

```
let commits = [];
let fetchFailed = false;
try {
  commits = fetchPlansCommits({ sinceDays: opts.sinceDays });
} catch (e) {
  console.error(`Could not read ${PLANS_REPO} commits: ${e.message}`);
  fetchFailed = true;
}
```

Then append `plans_fetch_failed=true` to `GITHUB_OUTPUT` so a failed fetch is visible to the caller and doesn't look like a clean no-op.

---

`.github/workflows/check-plans-sidebar-labels.yml` -- `npm install` → `npm ci`

`npm ci` is the right call in CI: faster, uses exact lock-file versions, and fails loudly if `package-lock.json` is out of sync.

---

`scripts/check-plans-sidebar-labels.js` -- `PRODUCTS_PATH` naming

`PRODUCTS_YAML` and `PRODUCTS_PATH` serve different purposes (absolute read path vs. prose display string) but are easy to conflate. `PRODUCTS_DISPLAY_PATH` would make the intent clear.

Piotr1215:
Thanks for the review. All three addressed in 5a21b472.

### 1. Silent failure in `fetchPlansCommits`

Tracked the failure separately and surfaced it. `main()` now sets a `fetchFailed` flag in the catch and appends `plans_fetch_failed=true` to `GITHUB_OUTPUT` on every output path (no-op, open-issue, report-only). A failed read of `loft-sh/plans` (for example a token without access) now reports `drift=false` + `plans_fetch_failed=true` instead of being indistinguishable from a genuine quiet week. The new output is documented in the header comment.

Verified with a failing `gh` shim plus `--force`: the error logs and `GITHUB_OUTPUT` reads `drift=true` / `plans_fetch_failed=true`.

### 2. `npm install` to `npm ci`

Done. `package-lock.json` is tracked, and the workflow already assumed a lockfile via `cache: 'npm'`, so this is safe and fails loudly on lockfile skew.

### 3. `PRODUCTS_PATH` naming

Renamed to `PRODUCTS_DISPLAY_PATH` (both occurrences), with a comment marking it as the prose display string versus the absolute read path `PRODUCTS_YAML`.

Test suite still green: 24/24 via `node --test tests/check-plans-sidebar-labels.test.js`.

djwfyi (changes requested):
Codex driven review has two suggestions I think are worth considering:

P2 [scripts/check-plans-sidebar-labels.js (line 241)](/Users/daryl/git/vcluster/vcluster-docs/scripts/check-plans-sidebar-labels.js:241) / [scripts/check-plans-sidebar-labels.js (line 410)](/Users/daryl/git/vcluster/vcluster-docs/scripts/check-plans-sidebar-labels.js:410)
The weekly trigger only looks back exactly 7 days from the current run time. If GitHub skips a scheduled run, delays it past the previous run’s timestamp, or the workflow is disabled briefly, loft-sh/plans commits in that gap are never reviewed. For something claiming to “check weekly for changes,” I’d use an overlap window, e.g. 8-14 days, or persist the last reviewed SHA/time.

P2 [scripts/check-plans-sidebar-labels.js (line 436)](/Users/daryl/git/vcluster/vcluster-docs/scripts/check-plans-sidebar-labels.js:436) / [scripts/check-plans-sidebar-labels.js (line 468)](/Users/daryl/git/vcluster/vcluster-docs/scripts/check-plans-sidebar-labels.js:468) / [.github/workflows/check-plans-sidebar-labels.yml (line 49)](/Users/daryl/git/vcluster/vcluster-docs/.github/workflows/check-plans-sidebar-labels.yml:49)
A failed loft-sh/plans fetch still exits green on scheduled runs when force=false and no commits were read. The script appends plans_fetch_failed=true, but the workflow step has no id and nothing consumes that output, so a bad token/private repo permission issue can silently become drift=false. I’d fail the job on scheduled fetch failure, or explicitly open/update an issue for the fetch failure.

Piotr1215:
Both Codex P2 items addressed in d051494b.

Lookback window: widened the default from 7 to 14 days (`DEFAULT_SINCE_DAYS`, ~2x the weekly cron) so a skipped, delayed, or briefly-disabled scheduled run no longer drops `loft-sh/plans` commits that land in the gap. The tracker is a single label-deduped issue, so the overlap updates the existing issue rather than spamming. Went with the overlap window over persisting a last-reviewed SHA to keep it stateless.

Failed fetch on scheduled runs: the script already emitted `plans_fetch_failed=true`, but the workflow step had no `id` and nothing consumed it, so a revoked `GH_ACCESS_TOKEN` still exited green as `drift=false` and looked like a quiet week. Gave the step `id: check` and added a follow-up step that fails the job when `steps.check.outputs.plans_fetch_failed == 'true'`, so lost access to `loft-sh/plans` turns the run red instead of silently skipping the review.

Verified with a failing `gh` shim on a non-forced (scheduled) run: `GITHUB_OUTPUT` gets `drift=false` + `plans_fetch_failed=true` and the new step exits 1. Test suite green (24/24, updated the `parseArgs` default test), actionlint clean.

djwfyi (approved):
LGTM

PR #2358, opened 2026-07-03
docs(ENGNODE-336): add information about wildcard in proxy config

matskiv:
# Content Description

Add information about wildcard(*) in API proxy config.

## Preview Link 

https://deploy-preview-2358--vcluster-docs-site.netlify.app/docs/vcluster/next/configure/vcluster-yaml/experimental/proxy#wildcard-kind-matching

## Internal Reference

Product change implemented in ENGNODE-336, no docs ticket.

@netlify /docs

### DIFF
--- a/vcluster/configure/vcluster-yaml/experimental/_code/proxy-wildcard-config.yaml
+++ b/vcluster/configure/vcluster-yaml/experimental/_code/proxy-wildcard-config.yaml
@@ -1,6 +1,6 @@
 # Wildcard Resource Proxy configuration
 # Proxies all resource kinds from the example.com/v1 API group to a target tenant cluster.
-# Use this when new Kinds are added to the API group frequently, avoiding config updates.
+# Use this when new kinds are added to the API group frequently, avoiding config updates.
 experimental:
   proxy:
     customResources:
--- a/vcluster/configure/vcluster-yaml/experimental/_code/proxy-wildcard-config.yaml
+++ b/vcluster/configure/vcluster-yaml/experimental/_code/proxy-wildcard-config.yaml
@@ -1,0 +1,10 @@
+# Wildcard Resource Proxy configuration
+# Proxies all resource kinds from the example.com/v1 API group to a target tenant cluster.
+# Use this when new Kinds are added to the API group frequently, avoiding config updates.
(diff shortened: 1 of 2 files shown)


djwfyi:
## Style review + accuracy check against upstream code

**Vale (0 errors, 3 warnings on CI, now 0/0):**
- Sentence-cased two headings (`Wildcard Kind matching` → `Wildcard kind matching`, and the matching example heading) — Google.Headings rule, and for consistency with the lowercase "kind" already used in the body prose.
- Replaced two `e.g.` with `for example` — Loft.no-latinisms rule.
- Lowercased "Kinds" → "kinds" in the new `proxy-wildcard-config.yaml` comment to match.

**Accuracy check against `vcluster-pro` (commits `5a4cd75d8` / `2b259bbae`, both `ENGNODE-336`):**

Traced the actual implementation (`pkg/config/validation.go`, `pkg/proxy/{proxy,handler}.go`) to verify the new prose before merging:

- Confirmed: `*` as the resource name in `*.group/version` is the real wildcard syntax, and it does route the entire group/version to the target with no kind-level filtering.
- Fixed: the mutual-exclusivity note said a wildcard and explicit key "cannot both be configured" for the same group/version. The validation code skips `enabled: false` entries before the conflict check (see the fix commit's message: "Disabled custom resource proxy entries do not produce a route at runtime ... so they should not participate in wildcard/explicit conflict ... checks"). So they can both exist in `vcluster.yaml` as long as the explicit one is disabled — only both being *enabled* is rejected. Reworded accordingly.
- Added: `handler.go`'s `serveAPIResources` deliberately returns an empty discovery list for wildcard proxies while the target is unreachable, to keep the APIService healthy, per the code comment "real discovery flows through once the target recovers." That's a real behavioral difference from explicit (non-wildcard) entries, which keep advertising their configured resource names during an outage. Documented this under Target unavailable errors so it doesn't surprise anyone running `kubectl api-resources` during a target outage.

Also removed a colon from the new troubleshooting prose to match our no-colons-in-prose house style.

djwfyi (approved):
Did a pass of my own with a few changes.
With those applied already, this looks good to me.

PR #2376, opened 2026-07-13
Codeblock colors updated to match vCluster Platform

adiron:
# Content Description

The colors of code blocks have been updated to match the vCluster Platform's terminal theme.

Additionally, the optional title of code blocks has been changed to the default fg color.

Before:
Screenshot 2026-07-13 at 09 58 35
After:
Screenshot 2026-07-13 at 09 58 32

Before:
Screenshot 2026-07-13 at 09 58 14
After:
Screenshot 2026-07-13 at 09 58 12

## Internal Reference

Closes DOC-

@netlify /docs

### DIFF
--- a/src/css/content/code.scss
+++ b/src/css/content/code.scss
@@ -2,6 +2,10 @@
   margin-top: var(--ifm-pre-padding);
 }
 
+[class*="codeBlockTitle"] {
+  color: #eceaea;
+}
+
 .docusaurus-highlight-code-line {
     background-color: rgba(255, 255, 255, 0.08);
     display: block;
--- a/src/prism-theme.js
+++ b/src/prism-theme.js
@@ -1,78 +1,80 @@
 /**
  * vCluster brand-aligned Prism syntax highlighting theme.
  *
- * Background: Black Pearl (#050B24)
- * Plain text: #CDD3DE (cool near-white, easy on the eyes)
+ * Background: #312727 (dark warm brown - from loft-enterprise Figma design tokens)
+ * Plain text: #eceaea (foreground - warm near-white default)
  *
- * Token palette — five deliberate choices:
- *   Sky blue     #48B5FF  strings / inserted diffs (matches vcluster.com highlight.js)
- *   Light orange #FFA366  numbers, booleans, variables (literal values)
- *   White        #FFFFFF  keywords, tags (structural / declarative tokens)
- *   Pale orange  #FFE0CC  YAML keys, bash commands, class names (brand family, readable)
- *   Blaze Orange #FF6600  shebangs, YAML anchors (rare, maximum signal)
- *   Slate       #7B8BB2  comments, punctuation, null (muted / de-emphasized)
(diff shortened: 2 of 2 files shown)


djwfyi:
## WCAG AA contrast review

I checked the proposed Prism colors against [WCAG 2.2 SC 1.4.3](https://www.w3.org/TR/WCAG22/#contrast-minimum), which requires at least 4.5:1 contrast for normal-sized text.

I also confirmed that these hex values come directly from the xterm theme in `loft-enterprise/ui/src/components/Terminal/Terminal.tsx`. Keeping the docs visually aligned with vCluster Platform makes sense. However, an ANSI terminal palette slot is not necessarily safe when it is repurposed as a syntax-token foreground.

### Proposed theme on the normal `#312727` background

All proposed colors pass except `brightBlack` (`#6e6a6a`), which is assigned to comments, punctuation, `null`, and `undefined`:

| Token color | Contrast | AA |
|---|---:|---|
| `#eceaea` | 12.09:1 | Pass |
| `#6e6a6a` | **2.71:1** | **Fail** |
| `#6eb2fe` | 6.53:1 | Pass |
| `#f9905c` | 6.33:1 | Pass |
| `#faf8f8` | 13.69:1 | Pass |
| `#b7b5b5` | 7.10:1 | Pass |
| `#f86f1f` | 5.02:1 | Pass |
| `#e47c72` | 5.12:1 | Pass |

### Highlighted lines

There is also an existing selector mismatch worth fixing here. Docusaurus 3.10.1 emits `theme-code-block-highlighted-line` and defaults its background to `#484d5b`. The custom `.docusaurus-highlight-code-line` rule in `code.scss` does not target that class.

Against the actual `#484d5b` highlighted-line background, these proposed colors fail AA: `#6e6a6a` (1.58:1), `#6eb2fe` (3.80:1), `#f9905c` (3.69:1), `#b7b5b5` (4.14:1), `#f86f1f` (2.93:1), and `#e47c72` (2.98:1).

The prior theme is compliant on ordinary code lines, but it also has failures on Docusaurus highlighted lines, so this is partly an existing issue rather than solely a regression in this PR.

### Recommendation while retaining the Platform scheme

1. Keep the Platform background, foreground, and the other proposed palette mappings.
2. Use the Platform `white` token (`#b7b5b5`) instead of `brightBlack` for comments, punctuation, `null`, and `undefined`. It remains visually muted relative to the foreground and gives 7.10:1 contrast on `#312727`.
3. Override `--docusaurus-highlighted-code-line-bg` to the Platform terminal `black` token (`#1e1313`) instead of using the Docusaurus gray. This also fixes the stale selector issue while staying entirely within the Platform terminal palette.

With those two changes, every mapped token passes AA: the lowest contrast is 5.02:1 on ordinary lines and 6.29:1 on highlighted lines.

If exact use of `brightBlack` for comments is a hard design requirement, that token itself needs an accessible lighter value for documentation text; there is no background-only adjustment that makes `#6e6a6a` pass while retaining the proposed normal background.

djwfyi:
Closing for now. Design may work on an alternative.

PR #2412, opened 2026-07-16
docs(DOC-1632): detect new upstream api kinds missing from platform docs generator

djwfyi:
# Content Description
Adds an ongoing check so the next batch of missing platform resource types (like the 8 fixed in DOC-1630) gets caught automatically instead of piling up silently.

- New tool: `hack/platform/check-resource-coverage` diffs every Kind registered in `loft-sh/api`'s management API scheme (parsed from `vendor/.../zz_generated.api.register.go`) against every `Resource:` field `hack/platform/partials/main.go` actually generates a reference page for. Exit codes 0/1/2 match the existing `hack/annotations` convention.
- New manifest: `hack/platform/check-resource-coverage/exclude.txt`, seeded with the 8 Kinds I could confidently classify as internal/action-style (RBAC self-checks like `SubjectAccessReview`, RPC-style action endpoints like `ConvertVirtualClusterConfig`). Deliberately does **not** include the other ~17 currently-uncovered Kinds (`AgentAuditEvent`, `Announcement`, `NodeType`, `NodeEnvironment`, `ResetAccessKey`, `Task`, etc.) — those haven't been individually vetted the way DOC-1630's 8 were, so the tool will correctly keep flagging them until someone does that triage (tracked separately, not blocking this PR).
- Wired into `handle-source-release.yml`'s `platform-released` path, right after the `loft-sh/api`/`agentapi` vendor bump (the step that makes a new Kind available in the first place). Non-fatal — emits a `::warning::` annotation rather than blocking the canonical-docs sync PR, since actually fixing a gap needs a human to write a realistic example object informed by real usage, the same investigation DOC-1630 required.

**Caught a bug while building it**: my first regex (`Resource:\s*"..."`) also matched `SubResource:` fields (substring match), inflating the covered count. Fixed by anchoring on line start (`(?m)^\s*Resource:`). Also found that my manual DOC-1630 investigation undercounted by 2 — `ResetAccessKey` and `Task` are genuinely registered resources I missed by hand that the tool now catches correctly.

**Verified**: ran locally against current `main` (17 covered, 25 gaps including DOC-1630's 8) and against DOC-1630's branch directly (25 covered, the 8 correctly drop out of the gap list, 17 remain for future triage).

## Preview Link
N/A — this PR only touches `hack/` tooling and CI workflow, no docs content.

## Internal Reference
Closes DOC-1632

@netlify /docs

### DIFF
--- a/.github/workflows/handle-source-release.yml
+++ b/.github/workflows/handle-source-release.yml
@@ -179,6 +179,27 @@ jobs:
           go mod tidy
           go mod vendor
 
+      # Checks whether the freshly-bumped loft-sh/api vendor tree registered
+      # any Kind that hack/platform/partials/main.go has no
+      # util.GenerateObjectOverview call for — the class of gap DOC-1630
+      # fixed for 8 pre-existing Kinds (see DOC-1632). Non-fatal: a missing
+      # Kind needs a human to write a realistic example object informed by
+      # how it's actually used, the same investigation DOC-1630 required, so
+      # this only surfaces the gap as a warning rather than blocking the
+      # canonical-docs sync PR or attempting to auto-fix it.
+      - name: Check platform resource generator coverage
+        if: steps.classify.outputs.skip != 'true' && steps.inputs.outputs.event == 'platform-released'
+        continue-on-error: true
+        run: |
+          set +e
+          go run ./hack/platform/check-resource-coverage --json-output coverage-report.json
+          set -e
+          gaps=$(jq -r '.gaps | length' coverage-report.json)
+          if [ "$gaps" -gt 0 ]; then
+            names=$(jq -r '.gaps | join(", ")' coverage-report.json)
+            echo "::warning::check-resource-coverage found $gaps registered API Kind(s) with no generator coverage: $names"
+          fi
+
       # vcluster + vcluster-cli generators consume the released vCluster
       # source tree. Platform generator builds against API types pinned in
       # this repo's go.mod (see bump step above), so no source checkout is
--- a/.github/workflows/handle-source-release.yml
+++ b/.github/workflows/handle-source-release.yml
@@ -192,11 +192,16 @@ jobs:
         continue-on-error: true
         run: |
           set +e
-          go run ./hack/platform/check-resource-coverage --json-output coverage-report.json
+          go run ./hack/platform/check-resource-coverage --json-output "${RUNNER_TEMP}/coverage-report.json"
+          status=$?
           set -e
-          gaps=$(jq -r '.gaps | length' coverage-report.json)
+          if [ ! -f "${RUNNER_TEMP}/coverage-report.json" ]; then
+            echo "::warning::check-resource-coverage exited $status without writing a report (tool may need updating, e.g. after a major API version bump); skipping coverage check"
+            exit 0
+          fi
+          gaps=$(jq -r '.gaps | length' "${RUNNER_TEMP}/coverage-report.json")
           if [ "$gaps" -gt 0 ]; then
-            names=$(jq -r '.gaps | join(", ")' coverage-report.json)
+            names=$(jq -r '.gaps | join(", ")' "${RUNNER_TEMP}/coverage-report.json")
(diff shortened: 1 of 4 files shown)


.github/workflows/handle-source-release.yml
```diff
@@ -179,6 +179,27 @@ jobs:
...
+      # Kind needs a human to write a realistic example object informed by
+      # how it's actually used, the same investigation DOC-1630 required, so
+      # this only surfaces the gap as a warning rather than blocking the
+      # canonical-docs sync PR or attempting to auto-fix it.
+      - name: Check platform resource generator coverage
+        if: steps.classify.outputs.skip != 'true' && steps.inputs.outputs.event == 'platform-released'
+        continue-on-error: true
+        run: |
+          set +e
+          go run ./hack/platform/check-resource-coverage --json-output coverage-report.json
```
Piotr1215: `--json-output coverage-report.json` writes to the repo root, and it isn't gitignored. The `Open or update PR` step runs `git add -A` in the same job (line 263), so this report gets committed into every docs-sync PR. Route it to `${RUNNER_TEMP}/coverage-report.json` here and in the two `jq` reads below, or gitignore it.

Piotr1215 (commented):
LGTM. Verified it builds and the counts match your write-up, go-pluralize is already a direct dep via schema.go so no go.mod churn, and the line-anchored `Resource:` regex correctly skips `SubResource:` and commented-out lines. A few non-blocking notes; the main one is the `coverage-report.json` artifact getting swept into the docs-sync PR.

.github/workflows/handle-source-release.yml
```diff
@@ -179,6 +179,27 @@ jobs:
...
+          set -e
+          gaps=$(jq -r '.gaps | length' coverage-report.json)
```
Piotr1215: On a tool error (exit 2, e.g. the hardcoded `v4` register path breaking after a major-version bump) the report file is never written, so this `jq` call fails and no `::warning::` fires. `continue-on-error` then hides it, so the checker goes silent exactly when it breaks, and right at a major bump when a new Kind is most likely. Maybe emit a warning when the report is missing.

hack/platform/check-resource-coverage/main.go
```diff
@@ -0,0 +1,149 @@
...
+	"strings"
+
+	pluralize "github.com/gertd/go-pluralize"
+)
+
+var (
+	registerPattern = regexp.MustCompile(`Management([A-Za-z]+)Storage\s*=\s*builders\.NewApiResourceWithStorage`)
+	// (?m)^\s*Resource: anchors on line start so this doesn't also match
+	// SubResource: fields, which contain "Resource:" as a substring.
+	generatorPattern = regexp.MustCompile(`(?m)^\s*Resource:\s*"([a-z0-9]+)"`)
```
Piotr1215, line 40: This regex already surfaced two real bugs (the `SubResource:` substring match and the undercount). A small table test would lock the line-start `Resource:` anchoring against regression. `hack/redirect-resolver` ships a `main_test.go` for the same reason.

.github/workflows/handle-source-release.yml
djwfyi: Fixed: the report now writes to `${RUNNER_TEMP}/coverage-report.json` (both here and in the two `jq` reads below), so it stays out of the working tree entirely and can't get picked up by `git add -A`.

.github/workflows/handle-source-release.yml
djwfyi: Fixed: the step now checks whether the report file exists after the tool exits. If it's missing, it emits `::warning::check-resource-coverage exited $status without writing a report...` and exits 0, instead of letting the `jq` call fail silently under `continue-on-error`.

hack/platform/check-resource-coverage/main.go
djwfyi: Added `main_test.go` with table tests against `generatorPattern` covering: a standalone `Resource:` field, a standalone `SubResource:` field (must not match), `Resource:` immediately followed by `SubResource:` on adjacent lines (must match only the former), and multiple `Resource:` fields across separate calls. This locks the line-start anchoring in against a regression.

PR #2418, opened 2026-07-17
docs(doc-1635): drop unsupported disruption budget reasons field

Piotr1215:
# Content Description

Align the auto nodes disruption budget example with the shipped vCluster values schema. The page documented a `reasons` field under `privateNodes.autoNodes.dynamic.disruption.budgets` that the schema does not accept, so the example, as written, fails validation.

## Key Changes

- Removed `reasons` from the disruption budget example in `vcluster/configure/vcluster-yaml/private-nodes/auto-nodes.mdx`.
- Added a one-line note that each budget supports the `nodes`, `schedule`, and `duration` fields.

## Evidence

Checked against `loft-sh/vcluster` (main, `76a1613f`):

- The `DynamicNodePoolDisruptionBudget` Go struct (`config/config.go`) exposes only `nodes`, `schedule`, and `duration`.
- The generated `chart/values.schema.json` marks that object `additionalProperties: false`, so a budget carrying `reasons` is rejected at config validation.
- `git log -S Reasons -- config/config.go` is empty: the field has never existed in the vCluster config.
- Upstream (vendored) Karpenter does expose `reasons` on its budget type, which is what the docs example had mirrored.

Verified drift-clean with the DOC-1626 `config-drift` checker (rebuilt from source): auto-nodes findings went from 2 to 0, total unknown findings 25 to 23, with no new findings introduced.

## Backporting

The same example appears in the released versions `0.28`, `0.33`, `0.34`, `0.35`, and `0.36`. Per repo convention, `vcluster_versioned_docs/` is not edited by hand and is backported automatically, so this PR touches only the `next` (top-level `vcluster/`) copy. The backport should carry this fix to those versions.

## Preview Link

- https://deploy-preview-2418--vcluster-docs-site.netlify.app/docs/vcluster/next/configure/vcluster-yaml/private-nodes/auto-nodes

## Internal Reference

Closes DOC-1635

@netlify /docs

### DIFF
--- a/vcluster/configure/vcluster-yaml/private-nodes/auto-nodes.mdx
+++ b/vcluster/configure/vcluster-yaml/private-nodes/auto-nodes.mdx
@@ -43,7 +43,7 @@ privateNodes:
     dynamic:
       - name: my-dynamic-node-pool
         limits:
-          nodes: 3
+          nodes: "3"
 ```
 
 :::info No vCluster restart required
@@ -65,27 +65,37 @@ privateNodes:
   enabled: true
   autoNodes:
     - provider: my-provider
-      properties:
-        # Exact match
-        my-propert: my-value
-
-        # One of
-        my-property/value-1: true
-        my-property/value-2: true
-        my-property/value-3: true
-
-        # Not in
-        my-property/value-1: false
-        my-property/value-2: false
-        my-property/value-3: false
-
-        # Exists
-        my-property: true
-
-        # Not exists
-        my-property: false
       static:
         - name: my-static-pool
+          quantity: 1
+          nodeTypeSelector:
+            # Exact match
+            - property: my-property
+              value: my-value
+
+            # One of
+            - property: my-property
+              operator: In
+              values:
+                - value-1
+                - value-2
+                - value-3
+
+            # Not in
+            - property: my-property
+              operator: NotIn
+              values:
+                - value-1
+                - value-2
+                - value-3
+
+            # Exists
+            - property: my-property
+              operator: Exists
+
+            # Not exists
+            - property: my-property
+              operator: NotExists
 ```
 
 The following operators are available and supported:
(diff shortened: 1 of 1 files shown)


djwfyi:
Verified the `reasons` fix independently against `config/config.go` and the generated schema (`additionalProperties: false` on `DynamicNodePoolDisruptionBudget`, `nodes`/`schedule`/`duration` only, no `reasons` in git history). That part is accurate.

While verifying, found that the same example (and 5 others on this page) still failed schema validation for an unrelated, older reason: `privateNodes.autoNodes` has been an array since vcluster#3388c395d ("refactor: auto nodes configuration", first released v0.30.0), but 6 of the 8 `autoNodes:` examples on this page still wrote it as a flat mapping instead of a list. Confirmed live against the vCluster YAML MCP validator, before this second fix the disruption-budget example still failed with `must be array` at `privateNodes/autoNodes` even with `reasons` removed.

Pushed a follow-up commit converting all remaining `autoNodes:` examples to the correct list form, so every example on the page now validates clean.

One more thing surfaced while re-validating that I did **not** touch, since it's out of scope for this PR: the "Node type selector" example has `static:` entries missing the required `quantity` field, and boolean-looking values under `properties` (`true`/`false`) that the schema expects as strings. The "Limits" example and the very first "static and dynamic node pools" example also pass `limits.cpu`/`limits.nodes` as bare integers where the schema requires strings. Flagging in case DOC-1626 wants a follow-up ticket for those.

djwfyi:
Follow-up: fixed the remaining three issues flagged above in the same branch, so the whole page is now schema-clean:

- Two `limits` examples set `cpu`/`nodes` as bare integers; both fields are string-typed in the schema (`additionalProperties: {type: string}` on `DynamicNodePool.limits`), so quoted them.
- The "Node type selector" example was structurally wrong, not just a typing issue. It used `autoNodes[].properties` (a flat key/value map intended for node-provider properties like region/subscription) with boolean values, but the operators the prose describes (`In`/`NotIn`/`Exists`/`NotExists`) belong to `nodeTypeSelector`, an array of `Requirement` objects (`property`/`operator`/`value`/`values`) nested under each `static`/`dynamic` pool entry. Rewrote the example to use `nodeTypeSelector` correctly and added the `quantity` field the `static` entry was missing.

All 8 YAML examples on the page now validate clean against the vCluster YAML MCP (main branch schema). No open issues left on this page as far as I can tell.

djwfyi (approved):

PR #2495, opened 2026-07-24
fix(release): skip docs-sync regen for platform lines older than the api pin

Piotr1215:
# Content Description

Fixes the `handle-source-release` receiver, which failed on `platform-released v4.10.6` ([run 30051361563](https://github.com/loft-sh/vcluster-docs/actions/runs/30051361563)) at the "Generate docs" step. This is a CI/release-automation change only; no docs prose is touched.

## Root cause

The platform partials generator (`hack/platform/partials/main.go`) reflects against the `loft-sh/api` Go types pinned in `go.mod`, and the receiver bumps that pin to the released version before running it. The generator source tracks the current api line: `Authentication`/`Connector`/`AuthenticationGithub` moved `management/v1` -> `storage/v1` in v4.11.0 (DEVOPS-1081, #2379) and `NodeProfile` is v4.11+ only (#2411). When an older line is released (`v4.10.6`), the receiver pins the api down to `v4.10.6`, where those types live elsewhere or do not exist, so the generator fails to compile:

```
undefined: storagev1.Authentication
undefined: storagev1.Connector
undefined: storagev1.AuthenticationGithub
undefined: managementv1.NodeProfile
undefined: managementv1.NodeProfileSpec
undefined: storagev1.NodeProfileSpec
```

One generator source cannot compile against both lines, because the types physically moved packages across the minor boundary. Fixing the forward direction (DEVOPS-1081) is what broke the backward one; that is the loop we keep hitting.

## Fix

`classify-version.sh` now treats platform regeneration as current-line-only: for `platform-released`, it skips (channel `stale-line`) when the released MAJOR.MINOR is older than the `loft-sh/api` pin read from `go.mod`. An older frozen platform minor is a snapshot whose api reference does not change on a patch release, so the receiver goes green and opens no PR instead of failing to compile. vcluster and vcluster-cli generators compile the source checked out at the released tag, so they still regenerate older lines and are untouched.

Verified against the real repo `go.mod` (api pinned at `v4.11.0-rc.6`):

| event | before | after |
|-------|--------|-------|
| `platform-released v4.10.6` | compile failure (red) | skip, `stale-line` (green) |
| `platform-released v4.9.3` | compile failure (red) | skip, `stale-line` (green) |
| `platform-released v4.11.0` | generate `version-4.11.0` | generate `version-4.11.0` |
| `platform-released v4.11.0-rc.7` | generate `version-4.11.0` | generate `version-4.11.0` |
| `platform-released v4.12.0` | current docs root | current docs root |

`classify-version.bats` extended with older-line skip and current-line generate coverage (25/25 pass); `shellcheck` clean; `run-generator.bats` unaffected.

## Internal Reference
Closes DEVOPS-1168

@netlify /docs

### DIFF
--- a/hack/release/classify-version.bats
+++ b/hack/release/classify-version.bats
@@ -49,6 +49,22 @@ EOF
 ]
 EOF
 
+    # go.mod pins the loft-sh/api line the platform generator compiles against.
+    # The fixture's current line is 4.9 (matches highest frozen), so a
+    # platform-released event on an older line (4.5-4.8) must classify as skip
+    # while a 4.9 patch still regenerates. classify-version reads only the
+    # api/v4 pin from this file for the platform stale-line guard.
+    cat >"$FIXTURE/go.mod" <<EOF
+module github.com/loft-sh/vcluster-docs
+
+go 1.26
+
+require (
+	github.com/loft-sh/agentapi/v4 v4.9.0
+	github.com/loft-sh/api/v4 v4.9.0
+)
+EOF
+
     export REPO_ROOT="$FIXTURE"
 }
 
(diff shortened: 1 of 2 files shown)


djwfyi:
Non-blocking portability note: the CI path runs on Ubuntu, so this doesn't affect the intended production workflow. However, the new `sed` expression in `classify-version.sh` uses GNU basic-regex `\+`. BSD `sed` on macOS doesn't recognize that quantifier, so it extracts no API pin, leaves `gen_key=-1`, and classifies a stale release such as `v4.10.6` with `skip=false`.

If we want the script and Bats suite to behave consistently for local macOS contributors, the expression could use portable extended regex syntax instead, for example `sed -E` with unescaped `+` and `(...)`. I don't consider this blocking given the Ubuntu-only runtime.

djwfyi (approved):
Added one non-blocking note, but otherwise LGTM.

Piotr1215:
Good catch, thanks. Fixed in bf0a7295d: switched to `sed -nE` with unescaped `+` and `(...)`, which both GNU and BSD sed accept, so the pin read and the older-line skip tests behave the same on macOS. Re-verified against the real `go.mod` (`v4.11.0-rc.6`): `v4.10.6` still skips (`stale-line`), `v4.11.x` still generates; bats 25/25, shellcheck clean.

PR #2502, opened 2026-07-27
docs: (DOC-1655) add CoreDNS scaling guidance for large node counts

djwfyi:
# Content Description
Documents CoreDNS's default single-replica setup, when it becomes a bottleneck at large node counts (especially AI/ML fleets), and how to scale it today via a fixed replica bump or a node-proportional autoscaler (cluster-proportional-autoscaler), plus what to monitor. Cross-linked from the CoreDNS config page.

## Preview Link
- [CoreDNS scaling - the decision and how to](https://deploy-preview-2502--vcluster-docs-site.netlify.app/docs/vcluster/next/deploy/control-plane/coredns-scaling)
- [CoreDNS bottlenecks - the explanation page](https://deploy-preview-2502--vcluster-docs-site.netlify.app/docs/vcluster/next/understand/coredns-bottlenecks)
- [CoreDNS timeouts - the incident response troubleshooting page](https://deploy-preview-2502--vcluster-docs-site.netlify.app/docs/vcluster/next/troubleshoot/coredns-dns-timeouts)

## Internal Reference
Closes DOC-1655

@netlify /docs

### DIFF
--- a/.vale.ini
+++ b/.vale.ini
@@ -125,3 +125,12 @@ BasedOnStyles =
 [CONTRIBUTING.md]
 Google.Headings = NO
 Loft.imperative-headings = NO
+
+# "etcd storage class" and "API server concurrency" are correct sentence case
+# (etcd is always lowercase per its own branding; API is a correctly-cased
+# acronym), but Google.Headings still flags both. Inline vale toggles around
+# these headings don't suppress the rule, so scope it off for the whole page
+# instead, matching the precedent in deploy-with-argocd.mdx.
+[**/deploy/control-plane/sizing-and-performance.mdx]
+BasedOnStyles = Loft, Google, Vale
+Google.Headings = NO
--- a/vcluster/configure/vcluster-yaml/control-plane/components/coredns.mdx
+++ b/vcluster/configure/vcluster-yaml/control-plane/components/coredns.mdx
@@ -59,6 +59,9 @@ The embedded CoreDNS feature allows you to run CoreDNS as part of the syncer, wh
 
 The embedded approach requires consideration of operational trade-offs. Troubleshooting becomes more complex since DNS and sync functions share the same container and process space. Updates and restarts affect both DNS and sync capabilities simultaneously, though all CoreDNS features remain supported.
 
+## Scale for large node counts
+
+By default, CoreDNS runs a single replica. This is sufficient for most tenant clusters, but large node counts, particularly AI and ML workloads with many pods per node, can exceed what one replica handles. For guidance on when to scale and how, see [Scale CoreDNS for large node counts](../../../../deploy/control-plane/sizing-and-performance.mdx#scale-coredns-for-large-node-counts) in the control plane sizing and performance guide.
 
 ## Example
 
--- a/vcluster/configure/vcluster-yaml/control-plane/components/coredns.mdx
+++ b/vcluster/configure/vcluster-yaml/control-plane/components/coredns.mdx
@@ -65,9 +65,12 @@ This applies to the separate deployment type (the default) only. With
 `controlPlane.coredns.embedded: true`, there's no `coredns` Deployment in
 `kube-system`: vCluster removes it entirely, so
 `controlPlane.coredns.deployment.replicas` and Deployment-level scaling don't
-apply. Embedded mode can't add independent CoreDNS replicas, but it can
-receive more capacity by raising `controlPlane.statefulSet.resources`, which
-resizes the whole control-plane pod rather than CoreDNS specifically.
+apply. Embedded mode can't add independent CoreDNS Deployment replicas, but
+two options still add DNS capacity. Raising `controlPlane.statefulSet.resources`
+resizes the whole control-plane pod rather than CoreDNS specifically, and
+raising `controlPlane.statefulSet.highAvailability.replicas` adds more active
+CoreDNS instances, since the `vc-kube-dns` Service points at the syncer pods
+when embedded is on.
 
 By default, CoreDNS runs a single replica. This is sufficient for most tenant
 clusters, but high query volume, particularly from AI and ML workloads with
--- a/vcluster/configure/vcluster-yaml/control-plane/components/coredns.mdx
+++ b/vcluster/configure/vcluster-yaml/control-plane/components/coredns.mdx
@@ -61,7 +61,20 @@ The embedded approach requires consideration of operational trade-offs. Troubles
 
 ## Scale for large node counts
 
-By default, CoreDNS runs a single replica. This is sufficient for most tenant clusters, but large node counts, particularly AI and ML workloads with many pods per node, can exceed what one replica handles. For guidance on when to scale and how, see [Scale CoreDNS for large node counts](../../../../deploy/control-plane/sizing-and-performance.mdx#scale-coredns-for-large-node-counts) in the control plane sizing and performance guide.
+This applies to the separate deployment type (the default) only. With
+`controlPlane.coredns.embedded: true`, there's no `coredns` Deployment in
+`kube-system`: vCluster removes it entirely, so
+`controlPlane.coredns.deployment.replicas` and Deployment-level scaling don't
+apply. There's currently no supported way to add DNS capacity in embedded
+mode.
+
+By default, CoreDNS runs a single replica. This is sufficient for most tenant
+clusters, but high query volume, particularly from AI and ML workloads with
+bursty pod scheduling, can exceed what one replica handles. For why that
+happens, see
+[Why does CoreDNS become a DNS bottleneck?](../../../../understand/coredns-bottlenecks.mdx).
+For how to scale it, see
+[Scale CoreDNS for large node counts](../../../../deploy/control-plane/coredns-scaling.mdx).
 
 ## Example
 
--- a/vcluster/configure/vcluster-yaml/control-plane/components/coredns.mdx
+++ b/vcluster/configure/vcluster-yaml/control-plane/components/coredns.mdx
@@ -69,7 +69,7 @@ apply. Embedded mode can't add independent CoreDNS Deployment replicas, but
 two options still add DNS capacity. Raising `controlPlane.statefulSet.resources`
 resizes the whole control-plane pod rather than CoreDNS specifically, and
 raising `controlPlane.statefulSet.highAvailability.replicas` adds more active
-CoreDNS instances, since the `vc-kube-dns` Service points at the syncer pods
+CoreDNS instances, since the `kube-dns` Service points at the syncer pods
 when embedded is on.
 
 By default, CoreDNS runs a single replica. This is sufficient for most tenant
--- a/vcluster/configure/vcluster-yaml/control-plane/components/coredns.mdx
+++ b/vcluster/configure/vcluster-yaml/control-plane/components/coredns.mdx
@@ -65,8 +65,9 @@ This applies to the separate deployment type (the default) only. With
 `controlPlane.coredns.embedded: true`, there's no `coredns` Deployment in
 `kube-system`: vCluster removes it entirely, so
 `controlPlane.coredns.deployment.replicas` and Deployment-level scaling don't
-apply. There's currently no supported way to add DNS capacity in embedded
-mode.
+apply. Embedded mode can't add independent CoreDNS replicas, but it can
+receive more capacity by raising `controlPlane.statefulSet.resources`, which
+resizes the whole control-plane pod rather than CoreDNS specifically.
 
 By default, CoreDNS runs a single replica. This is sufficient for most tenant
 clusters, but high query volume, particularly from AI and ML workloads with
--- a/vcluster/deploy/control-plane/benchmark-methodology.mdx
+++ b/vcluster/deploy/control-plane/benchmark-methodology.mdx
@@ -1,7 +1,7 @@
 ---
 title: Control plane benchmark methodology
 sidebar_label: Benchmark methodology
-sidebar_position: 1
+sidebar_position: 2
 description: Test setup, service level indicators, and analysis for vCluster control plane scalability benchmarks.
 ---
 
--- a/vcluster/deploy/control-plane/binary/_category_.json
+++ b/vcluster/deploy/control-plane/binary/_category_.json
@@ -1,5 +1,5 @@
 {
   "label": "As a binary",
-  "position": "3",
+  "position": "6",
   "className": "standalone private-nodes"
 }
--- a/vcluster/deploy/control-plane/coredns-scaling.mdx
+++ b/vcluster/deploy/control-plane/coredns-scaling.mdx
@@ -49,13 +49,13 @@ controlPlane:
       replicas: 3
 ```
 
-vCluster re-applies the CoreDNS Deployment, including `spec.replicas`, every
-time the control plane pod starts, using a client-side apply that overwrites
-the live object. A restart, upgrade, or leader change resets `replicas` back
-to whatever `vcluster.yaml` specifies. Set `replicas` in `vcluster.yaml` to the
-floor you want rather than scaling the Deployment directly. The same
-overwrite-on-restart behavior applies to any other CoreDNS setting changed
-out-of-band. See [Configuration Persistence](../../configure/vcluster-yaml/control-plane/components/coredns.mdx#deploy-separate-coredns) for details.
+vCluster re-applies the CoreDNS Deployment every time the control plane pod
+starts, reconciling it against `vcluster.yaml` the same way `kubectl apply`
+reconciles a manifest against a live object. `replicas`, and any other
+CoreDNS setting, converges to whatever `vcluster.yaml` specifies rather than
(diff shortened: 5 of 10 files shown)


djwfyi:
## How the estimated numbers were derived

The two `Needs confirmation from the Control Plane team` callouts in this PR contain numbers that are reasoned estimates, not measurements from vCluster's own benchmark methodology (KWOK + kube-burner, as used elsewhere on this page). Flagging the reasoning so engineering can sanity-check or replace them:

**"When it becomes a bottleneck" (node count / QPS)**
- vCluster's default CoreDNS resource profile caps each replica at 1 CPU (`controlPlane.coredns.deployment.resources.limits.cpu: 1000m`).
- Published CoreDNS benchmarks put the CPU-bound cache-miss path (the `kubernetes` and `forward` plugins, which is what dominates under high pod churn) at roughly 2,000-3,000 sustained QPS per core.
- The historical `kube-dns-autoscaler` convention (predecessor to `cluster-proportional-autoscaler`) defaulted to about 1 replica per 16 nodes for general-purpose clusters.
- The reporting customer's own working ratio (1:8) is roughly double that density, which tracks with AI/ML fleets having higher pod density per node and more cache-miss churn, plus `ndots:5` search-domain expansion multiplying real query volume per application-level lookup.
- From those anchors, the doc estimates ~16-30 nodes for general-purpose Tenant Clusters, tightening to ~8-16 nodes for AI/ML fleets, at ~2,000-3,000 QPS/replica.

**"What to watch" table (DNS latency / error rate alert thresholds)**
- Left as `Needs confirmation` with no estimate. I don't have a defensible external anchor for CoreDNS's own P99 latency or SERVFAIL/NXDOMAIN rate under vCluster's specific setup the way I do for the QPS/node-count numbers above, so this needs real data rather than an estimate.

**Autoscaler mechanism**
- The doc recommends `cluster-proportional-autoscaler` as the tool for the node-proportional pattern (it's the standard/only real off-the-shelf option, and the same mechanism most managed Kubernetes distros use for their own CoreDNS). This part is verified against vCluster's source: the `coredns` Deployment in `kube-system` is a native object inside the Tenant Cluster (confirmed in `pkg/coredns/template.go` / `pkg/setup/controllers.go`), not a synced mirror, so the autoscaler can run entirely inside the tenant cluster with no cross-cluster kubeconfig split.
- Not verified: whether this is literally what the customer deployed, or whether they used a custom script. The `nodesPerReplica: 8` config value in the example reproduces their stated ratio either way, but if their actual implementation differs materially, worth calling out.

## What engineering needs to confirm before this ships

1. Real node-count and DNS QPS thresholds where a single CoreDNS replica saturates, ideally from the existing control-plane load-testing framework or from the reporting customer's own reported numbers.
2. DNS query P99 latency and error-rate (SERVFAIL/NXDOMAIN) alert thresholds for the "what to watch" table — currently has no estimate at all.
3. Whether the customer's actual autoscaling setup matches the `cluster-proportional-autoscaler` example in this PR, or differs in a way worth reflecting.

---
**Update:** applied fixes for the four issues raised in review — bumped the autoscaler image to v1.10.3, clarified `preventSinglePointFailure`'s 2-replica floor, dropped `NXDOMAIN` from the generic error signal (it's a normal byproduct of `ndots:5` search-path expansion, not an overload signal), and added a metric-expression/window/workload-input/SLO table under "What to watch" so engineering has a concrete framework rather than prose placeholders. Also flagged that confirming the numbers needs a new DNS workload in the load-testing framework, not a re-read of existing KWOK/kube-burner runs, since those don't model DNS load today.

djwfyi:
Staying in draft until we resolve all of the "Need confirmation" sections.

vcluster-dv:
Thanks for flagging the estimates yourself instead of quietly shipping them. I checked the code and chased where our numbers came from.

**Confirmed:** one replica by default and the 1000m CPU limit (both in `chart/values.yaml` under `controlPlane.coredns.deployment`, cite that and not `values.schema.json`, which has no defaults). The `coredns` Deployment really is a native object in the tenant cluster and its pods sync to the Control Plane Cluster. The autoscaler manifest itself is clean: image tag, flags, ConfigMap keys, target and RBAC all check out against cluster-proportional-autoscaler v1.10.3, it copy-pastes and starts. The PromQL works with no extra setup, since the default Corefile enables the prometheus plugin on :9153. And you are right that our load-test framework has no DNS workload today. The `ndots:5` point is the best thing on the page.

## Needs changing

**Embedded CoreDNS is the headline.** With `controlPlane.coredns.embedded: true` there is no `coredns` Deployment in `kube-system` at all, vCluster actively deletes it. So the autoscaler has no scale target, and `controlPlane.coredns.deployment.replicas` is silently ignored: no validation error, no warning, no extra DNS capacity. Placement makes it worse, the cross-link you added sits right under "Deploy embedded CoreDNS", so embedded readers hit this first. Needs a warning at the top of the section and the cross-link scoped to the separate deployment. I do not yet know the supported way to add DNS capacity in embedded mode, so treat that as an open question rather than documenting HA replicas as the answer.

Four behavior fixes:

- **Replicas do not stick.** vCluster re-applies the CoreDNS Deployment on every control plane start including `spec.replicas`, client-side apply with `Overwrite: true`. An autoscaler value is reset on restart, upgrade or leader change. Tell readers to set `replicas` to the floor they want, and link the Configuration Persistence note that is already on `coredns.mdx`.
- **The default topology spread constraint** (maxSkew 1 over `kubernetes.io/hostname`, `DoNotSchedule`) allows one CoreDNS pod per node. Replicas above the schedulable node count stay Pending.
- **The node count the autoscaler sees is not the fleet.** With the default `sync.fromHost.nodes.enabled: false` the tenant cluster only has node objects for nodes running its own pods, so the ratio lags placement. Also warn against `coresPerReplica` in that mode, synthesized nodes report a fixed 16 CPU allocatable.
- **`preventSinglePointFailure` does not override `min`.** It raises the node-derived count to 2 once there is more than one schedulable Ready node, and can override `max` if `max` is below 2.

One line each also worth adding: the 170Mi per-replica memory limit, because an OOMKilled CoreDNS looks identical to a saturated one and more replicas do not raise that cap; and `high-availability-k8s.mdx`, which already tells people to set CoreDNS to 3 replicas, so the two pages should not contradict each other.

## Has to be settled before merge

The numbers, and this is the part I cannot sign off.

- 2,000-3,000 qps per core does not match anything published. Upstream benchmarked these same plugin paths at roughly 17,700 backend qps per core. A figure 6x low makes us look an order of magnitude weaker than we are.
- The 16-30 and 8-16 node bands inherit that error, and node count is the wrong variable anyway. Load is pods times lookups per pod times search expansion.
- The 1:8 ratio is real but it is one account, not "AI Cloud accounts": Boost Run, Pylon #6962. They run their own CoreDNS via Helm with `controlPlane.coredns.enabled: false` at a 100m CPU limit, and picked 8 to neutralize core-based scaling on very large bare metal nodes. It supports neither the bundled-replica framing nor the 1 CPU ceiling.
- The ">8 nodes" figure that probably inspired the AI band belongs to the konnectivity agent, not CoreDNS. That conflation was untangled on 24 Jul, let us not put it back in customer-facing docs.
- Our only real observation points the other way: in ENGCP-967 a single default replica served an 8-worker standalone HA cluster for 21 days, and the DNS failure there was a kine watch stall.

On publishing the literal words "Needs confirmation": I do not think it holds. People copy numbers out of sizing pages and leave the caveat behind, and search and LLM indexing strip it entirely. I would cut the qps figure, both node bands, and the SLO rows we cannot source, and keep the qualitative guidance, which is the useful 80% here. Give upstream's convention as the starting point instead (one replica per 16 nodes and per 256 cores) and label 1:8 as a deliberate tightening. Replace the node bands with the query-rate method and a worked example so people can measure their own fleet. To be blunt on those bands: they should not stay and should not be widened either, because widening still reads as measured.

Two more. The GPU premise is inverted, GPU nodes usually run fewer pods per node and steady-state training is quiet on DNS. The real pressure is the job-launch burst, which is a better reason to scale and still supports your conclusion. And soften the NXDOMAIN line, sustained NXDOMAIN for Services the API server still returns is exactly how we caught ENGCP-967, so it is a real signal, just not a capacity one.

## The ask

One decision from you: cut the unsourced numbers and ship the rest now, or hold the page until we have a DNS load test. I recommend the first, I do not want to block your page on a benchmark that does not exist yet.

On my side: I will get the embedded-mode answer from the Pro side and post wording for the caveat here, get the autoscaler recipe actually exercised against a tenant cluster before we word it as validated, and keep the numbers tracked on ENGCP-1043 (Product Backlog, no committed release, so please do not phrase anything as an upcoming feature).

Small stuff, take or leave: three lowercase "control plane cluster" on a page that title-cases it elsewhere; "AI Cloud and ML workloads" uses AI Cloud as a customer category to describe a workload type, "AI and ML workloads" is what you want there; and the autoscaler pod spec has no securityContext, so it gets rejected in tenant clusters enforcing restricted Pod Security.

djwfyi:
Addressed the review feedback, and along the way the page outgrew its original spot, so it's now split into three pages.

## How @vcluster-dv's feedback was addressed

**Embedded CoreDNS**
- Added an explicit warning that none of the scaling guidance applies to embedded CoreDNS (`controlPlane.coredns.embedded: true`): there's no `coredns` Deployment to act on, and `replicas`/the autoscaler are silently ignored.
- Explained why: embedded CoreDNS runs inside the syncer container, not as its own Deployment (it's also a Pro feature).
- Added the one workaround that does exist: raising `controlPlane.statefulSet.resources` resizes the whole control-plane pod (a blunt instrument, but real). Confirmed this is a live, no-redeploy change, distinct from the embedded/separate deployment *type*, which genuinely is fixed at deploy time (per the `Predeployment configuration options` lists in `binary/basics.mdx` / `kubernetes-pod/basics.mdx`).
- Left `high-availability-k8s.mdx` unlinked: traced it down and it's an orphaned partial (moved to `docs/_partials/` in a "clean unused partial files" commit), not imported or published anywhere, and it uses the pre-`controlPlane.*` schema. Linking to it would point readers at dead, schema-stale content, so no cross-link was added.

**Replica behavior**
- Documented that vCluster re-applies the CoreDNS Deployment (including `spec.replicas`) via client-side apply on every control-plane restart/upgrade/leader change, so `replicas` changes need to go through `vcluster.yaml`, not a direct Deployment edit.
- Documented the default topology spread constraint (`maxSkew: 1` over `kubernetes.io/hostname`, `DoNotSchedule`): replicas above the schedulable node count stay `Pending`.
- Corrected `preventSinglePointFailure`: it raises the node-derived floor to 2 once there's more than one schedulable Ready node, it doesn't override `min`.
- Added a "Settings to avoid" note on `coresPerReplica`: with the default `sync.fromHost.nodes.enabled: false`, synthesized Node objects report a fixed 16 CPU allocatable, which breaks core-based scaling, and the autoscaler's node count lags real fleet size.
- Added the 170Mi default memory limit as its own monitoring signal, since an OOMKilled replica looks identical to a CPU-saturated one but isn't fixed by adding replicas.
- Added a `securityContext` to the autoscaler pod spec so it doesn't get rejected under restricted Pod Security.

**Unsourced numbers**
- Cut the 2,000-3,000 qps/core figure and both node-count bands (16-30 / 8-16) entirely, along with the `Needs confirmation` caution box wording, per your recommendation to ship the qualitative guidance now rather than gate on a benchmark that doesn't exist yet.
- Replaced the node-band guidance with a query-rate measurement method (compare `coredns_dns_requests_total` against replica CPU saturation) so readers can find their own ceiling instead of borrowing ours.
- Reframed the 1:8 ratio as one deployment's tightening (very large bare-metal GPU nodes, high per-node core count) rather than "AI Cloud accounts," and set the documented default to upstream's own convention (`nodesPerReplica: 16`).
- Fixed the GPU premise: GPU nodes typically run *fewer* pods per node, the actual pressure point is the job-launch burst, not steady-state density.
- Softened the `NXDOMAIN` framing: sustained `NXDOMAIN` for a resolvable Service is a real signal, but it's not a standalone capacity trigger on its own, since `ndots:5` expansion produces it as normal background noise too.

**Small stuff**
- "AI Cloud and ML workloads" → "AI and ML workloads." Fixed.
- The lowercase "control plane cluster" instances: left as lowercase rather than made consistent with the page's other title-cased instances, since lowercase is the current house style for this term. That's a pre-existing inconsistency across the page unrelated to this PR, so didn't do a wider sweep.

## Why it's now three pages instead of one

The CoreDNS section had grown to roughly half the original page and didn't share its methodology: everything else on `sizing-and-performance.mdx` is benchmarked against the KWOK/kube-burner control-plane framework, while CoreDNS explicitly can't be (hence the page needing its own disclaimer note just to justify being there).

Splitting by intent, not just by size:
- **[`understand/coredns-bottlenecks.mdx`](https://github.com/loft-sh/vcluster-docs/blob/DOC-1655/coredns-scaling/vcluster/understand/coredns-bottlenecks.mdx)** — the explanation: why query volume beats node count, the GPU/job-burst nuance, how to read `SERVFAIL`/`NXDOMAIN`, why embedded can't scale.
- **[`deploy/control-plane/coredns-scaling.mdx`](https://github.com/loft-sh/vcluster-docs/blob/DOC-1655/coredns-scaling/vcluster/deploy/control-plane/coredns-scaling.mdx)** — the how-to: the decision guide, both scaling methods, and monitoring setup. This is what stayed linked from `sizing-and-performance.mdx`.
- **[`troubleshoot/coredns-dns-timeouts.mdx`](https://github.com/loft-sh/vcluster-docs/blob/DOC-1655/coredns-scaling/vcluster/troubleshoot/coredns-dns-timeouts.mdx)** — new, symptom-first troubleshooting page for someone with an active incident rather than someone doing capacity planning. All three cross-link to each other.

Let me know if you'd like a second pass on any of these before merge.

vcluster-dv:
You took all of it, and the three-page split reads better than the single section. Thanks for turning it around that fast.

One thing to walk back before merge. I flagged the embedded lever as an open question, and the page now presents it as settled in three spots, strongest as "the only lever available". Please soften that to "the lever we know of today" until I confirm two things: whether raising the control-plane pod's resources actually gives embedded CoreDNS more headroom, and whether `controlPlane.statefulSet.highAvailability.replicas` also multiplies DNS serving, which would make "only" wrong.

Two other new claims worth a second source: that embedded CoreDNS is a Pro feature, and that the deployment type can't be switched on a live tenant cluster without redeploying. The second one appears twice and customers will act on it.

Asking the team about the embedded internals today, will post the wording here.

vcluster-dv:
Got the answers from Catalin, who did the embedded work in ENGCP-890. Both claims I flagged need to change.

Embedded components run as child processes of the syncer, in the same pod. So:

- Raising `controlPlane.statefulSet.resources` does help. That part is right.
- `controlPlane.statefulSet.highAvailability.replicas` also multiplies DNS serving. More syncer replicas means more active CoreDNS instances. With embedded off, `vc-kube-dns` points at the CoreDNS deployment pods; with it on, the service points at the syncer pods on TCP/UDP 1053. So "the only lever available" is wrong. Please drop "only" and list both.
- The deployment type is not fixed at deploy time. The flag can be flipped on a live tenant cluster, Catalin has done it. He is double checking whether there is a brief DNS gap during the switch, so word it as switchable and I will follow up if we need a downtime caveat.

Still open: the Pro-feature claim for embedded CoreDNS. No second source yet.

djwfyi:
Pushed fixes for both open items from Catalin's answers.

**Levers, not "only"** — dropped "the only lever available" everywhere it appeared (`coredns-bottlenecks.mdx`, `coredns-dns-timeouts.mdx`, `coredns.mdx`). Now documents both: raising `controlPlane.statefulSet.resources` (resizes the whole control-plane pod) and raising `controlPlane.statefulSet.highAvailability.replicas` (more syncer replicas means more active embedded CoreDNS instances, since `vc-kube-dns` points at the syncer pods on TCP/UDP 1053 when embedded is on).

**Switchable, not fixed at deploy time** — reversed that claim in both places it appeared. Kept it factual rather than narrating the open question: "You may see a brief DNS gap during the switch," no mention of it still being confirmed internally.

**Pro-feature claim** — checked `src/data/features.yaml` and found `vcp-distro-built-in-coredns` ("Built-In CoreDNS") listed under the `vCluster Pro Distro` category, so that's an independent second source. Left it as-is.

Also tightened "runs inside the syncer container" to your more precise "child processes of the syncer, in the same pod," and swapped "levers" for "options" throughout per a style pass.

Pushed to `DOC-1655/coredns-scaling` at 2e1cb7b.

vcluster-dv (approved):
Approving, with the caveat that this is a provisional sign off since the PR is still in draft. If anything material changes before you take it out of draft, re-ping me and I will look again.

On my side there is nothing left blocking. The embedded CoreDNS wording is fixed, and the page no longer claims the autoscaler recipe is validated, so I am not holding the merge on me exercising it first. If we do run it later and the numbers disagree with the page, we update the page then.

vcluster-dv (approved):

PR #2472, opened 2026-07-20
docs(observability): document GPU monitoring templates and NVSentinel

flomedja:
# Content Description

## Preview Link 

- [Overview - updated](https://deploy-preview-2472--vcluster-docs-site.netlify.app/docs/platform/next/maintenance/observability/overview)
- [Quickstart - NEW](https://deploy-preview-2472--vcluster-docs-site.netlify.app/docs/platform/next/maintenance/observability/quickstart)
- [GPU Monitoring Templates - NEW](https://deploy-preview-2472--vcluster-docs-site.netlify.app/docs/platform/next/maintenance/observability/gpu-monitoring-templates)
- [NVSentinel monitoring - NEW](https://deploy-preview-2472--vcluster-docs-site.netlify.app/docs/platform/next/maintenance/observability/nvsentinel-gpu-monitoring)

Note: Every page in the observability nav section has some updates. The above pages are the main ones.

## Internal Reference

Closes ENGPLAT-834
Closes DOC-1615

AI review: mention `@claude` in a comment to request a review or changes. See [CONTRIBUTING.md](https://github.com/loft-sh/vcluster-docs/blob/main/CONTRIBUTING.md#ai-assisted-pr-review) for available commands.

> FORK LIMITATION: `@claude` does not work on pull requests opened from forks. GitHub Actions cannot access the required secrets for fork-originated PRs. To use AI review, push your branch directly to this repository.

@netlify /docs

PR #2519, opened 2026-07-30
docs(platform/observability): requests/limits for pods

seanschneeweiss:
# Content Description

Add new resources parameter for Grafana, Prometheus and OTEL Collector, introduced by loft-sh/loft-enterprise#7617

## Preview Link 

https://deploy-preview-2519--vcluster-docs-site.netlify.app/docs/platform/next/maintenance/observability/install-observability-gateway#deploy-the-metrics-backend-with-argo-cd
https://deploy-preview-2519--vcluster-docs-site.netlify.app/docs/platform/next/maintenance/observability/troubleshooting#observability-pods-are-oomkilled-or-evicted

## Internal Reference

Part of ENGPLAT-798

@netlify /docs

### DIFF
--- a/platform/maintenance/observability/troubleshooting.mdx
+++ b/platform/maintenance/observability/troubleshooting.mdx
@@ -134,13 +134,12 @@ kubectl describe pod -n vcluster-platform <pod> | grep -A3 "Last State"
 If you see `OOMKilled` or `Evicted`, set resource requests and limits on the affected
 template:
 
-- Prometheus: the `resources` parameter. See [Set Prometheus resource requests and
-  limits](./install-observability-gateway.mdx#set-prometheus-resource-requests-and-limits).
-- Grafana: the `resources` parameter. See [Set Grafana resource requests and
-  limits](./query-fleet-metrics.mdx#set-grafana-resource-requests-and-limits).
-- Collectors: the `otelDaemonResources` and `otelDeployResources` parameters. See [Set
-  collector resource requests and
-  limits](./configure-edge-collectors.mdx#set-collector-resource-requests-and-limits).
+- Prometheus: the `resources` parameter. See [Deploy the metrics backend with Argo
+  CD](./install-observability-gateway.mdx#deploy-the-metrics-backend-with-argo-cd).
+- Grafana: the `resources` parameter. See [Deploy the bundled Grafana with Argo
+  CD](./query-fleet-metrics.mdx#deploy-the-bundled-grafana-with-argo-cd).
+- Collectors: the `otelDaemonResources` and `otelDeployResources` parameters. See [Deploy
+  collectors with Argo CD](./configure-edge-collectors.mdx#deploy-collectors-with-argo-cd).
 
 Prometheus is the most likely to be OOMKilled, because its memory grows with active series
 count.
--- a/platform/maintenance/observability/troubleshooting.mdx
+++ b/platform/maintenance/observability/troubleshooting.mdx
@@ -136,11 +136,27 @@ kubectl get pod -n vcluster-platform <pod> -o jsonpath='{.status.message}'
 The fix depends on which failure you see:
 
 - `OOMKilled`: the container exceeded its memory limit, or the node ran out of memory.
-  Measure actual usage with `kubectl top pod -n vcluster-platform` and set or raise the
-  memory limit accordingly. A limit that is too low causes this on its own.
-- `Evicted` under memory or CPU pressure: set memory and CPU requests so the kubelet
-  ranks the pod above pods that exceed their requests. The status message names the
-  pressure signal that triggered the eviction.
+  `kubectl top pod -n vcluster-platform <pod> --containers` only shows current per-container
+  usage. A restarted container resets to a fresh baseline, so it won't show the memory spike
+  that preceded the restart. Check historical usage instead. The bundled [golden signals
+  queries](../monitoring/fleet-monitoring-otel.mdx#golden-signals-queries) filter on tenant
+  workload labels that these platform pods don't carry, so query the platform namespace
+  directly in the bundled Prometheus or Grafana. Scope the query to the specific container
+  the resource parameter targets, since a pod can run more than one:
+
+  ```promql
+  max(max_over_time(container_memory_working_set_bytes{namespace="vcluster-platform", pod="<pod>", container="<container>"}[1h]))
+  ```
+
+  Use the peak this returns to size a limit that covers it. A limit that is too low causes
+  this on its own.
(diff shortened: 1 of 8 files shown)


djwfyi (commented):
I found two Kubernetes behavior issues in the new troubleshooting guidance:

platform/maintenance/observability/troubleshooting.mdx
```diff
@@ -118,6 +118,32 @@ can't connect:
 - If the gateway uses a `LoadBalancer` Service and a Platform-managed certificate, wait for
   the LoadBalancer address before expecting the gateway Deployment to become available.
 
+## Observability pods are OOMKilled or evicted
+
+The bundled templates deploy Prometheus, Grafana, and the collectors without an explicit
+`resources` block, so those containers get the `BestEffort` QoS class and the kubelet
```
djwfyi: **[P2] Qualify the `BestEffort` claim**

Kubernetes assigns QoS to the pod, not individual containers, and a pod is `BestEffort` only when every container lacks CPU and memory settings and admission has not supplied defaults through a `LimitRange`. Eviction ordering also depends on usage relative to requests and pod priority; QoS does not govern every type of node pressure. Could we phrase this as the template defaults making these pods *likely* `BestEffort` and recommend checking `.status.qosClass`?

platform/maintenance/observability/troubleshooting.mdx
```diff
@@ -118,6 +118,32 @@ can't connect:
...
+evicts them first under node pressure.
+
+Check the restart reason:
+
+```bash
+kubectl get pods -n vcluster-platform -o wide
+kubectl describe pod -n vcluster-platform <pod> | grep -A3 "Last State"
+```
+
+If you see `OOMKilled` or `Evicted`, set resource requests and limits on the affected
```
djwfyi: **[P2] Distinguish OOM kills from eviction**

This prescribes requests and limits for two different failure modes. A memory limit can itself cause `OOMKilled`, while an `Evicted` pod may have been removed for disk or PID pressure, which CPU and memory settings will not fix. Please have users inspect the terminated-container and pod eviction messages first; for a limit OOM, measure usage and raise the memory limit, and for memory-pressure eviction, set an appropriate request.

platform/maintenance/observability/troubleshooting.mdx
seanschneeweiss: The likely wording is a good hint.

platform/maintenance/observability/troubleshooting.mdx
seanschneeweiss: Agree, that is helping users more to find the root cause.

djwfyi:
Reverted the terminate → delete wording change in `nvsentinel-gpu-observability.mdx` (commit a39be3fb1).

**Why:** the docs already draw a line elsewhere between the two verbs — "delete" for a `kubectl delete` of a Kubernetes API object (matching the literal command and the `Terminating` status field), and "terminate" for a controller destroying the underlying cloud instance/VM (matching the actual provider API names, e.g. `ec2:TerminateInstances`, which the AWS Janitor row still references a few lines away). The Janitor's action is the second case — it calls the cloud provider directly, not `kubectl delete node` — so "terminate" is the more accurate and consistent term here.

**Root cause of the original Vale warnings:** turns out there's no real style conflict to work around. `.vale.ini` already had a `Google.WordList = NO` override for this exact false positive, with a comment explaining the same distinction above. But it was scoped to `nvsentinel-gpu-monitoring.mdx` — the file's name before it got renamed to `nvsentinel-gpu-observability.mdx` in the same PR (#2472) that added the override. The glob went stale and silently stopped matching, so Vale started flagging "terminate" again with no obvious cause. Fixed the glob to point at the current filename, confirmed Vale is clean on the reverted wording.

Also added a short section to the docs-writer skill's style guide documenting the delete-vs-terminate distinction and this stale-glob failure mode, so it's easier to catch next time a page gets renamed.

djwfyi:
Addressed Codex's follow-up findings on the OOMKilled/evicted troubleshooting section (commit bf418d152):

**CPU pressure doesn't evict pods.** Fixed — kubelet node-pressure eviction signals are memory, disk/inode, and PID availability, not CPU. CPU contention is throttled, not evicted, so a CPU request can't prevent an eviction. Added a note that `DiskPressure`/`PIDPressure` evictions need the underlying node condition resolved, not a requests/limits change.

**`kubectl top` can't size the failed limit.** Fixed — switched to `kubectl top pod <pod> --containers` for per-container usage, and noted a restarted container resets to a fresh baseline so it won't show the spike that caused the OOM kill in the first place.

**Linked query excluded the affected pods.** Also fixed — the golden-signals query I originally linked filters on `vcluster_name`/`vcluster_virtual_pod` labels that only tenant workloads carry, so it would've returned nothing for the platform-namespace Prometheus/Grafana/collector pods this section is about. Replaced it with a `max_over_time(...)` query scoped directly to the platform namespace and the specific container the resource parameter targets (since a pod can run more than one), wrapped in `max()` for a single peak value.

**`compute.instances.reset` isn't termination.** Fixed in the style guide's word-choice section — `compute.instances.reset` only reboots a GCP VM without destroying it, so it shouldn't sit next to `ec2:TerminateInstances` as a "terminate" example. Split reboot-only APIs out explicitly so this doesn't get miscited as termination in future docs.

**`Terminating` isn't a `status.phase` value.** Also fixed — corrected the guide to describe it as a kubectl-displayed state driven by `metadata.deletionTimestamp`, not an actual Pod phase enum value (those are `Pending`/`Running`/`Succeeded`/`Failed`/`Unknown`).

Vale, link, and whitespace checks all pass on the current revision.

seanschneeweiss:
@djwfyi 
thanks for all the changes and fixing vale.

seanschneeweiss:
The troubleshooting for **Linked query excluded the affected pods** is probably a bit too detailed. Generally OOM kills are easy to catch, right? And if Prometheus is crashlooping, it is not really possible to access the metrics and run the PromQL query against it.
If we could make this a minimal troubleshooting guide for the crashlooping and OOM killed case, I think that would be sufficient. WDYT?

djwfyi:
> The troubleshooting for **Linked query excluded the affected pods** is probably a bit too detailed. Generally OOM kills are easy to catch, right? And if Prometheus is crashlooping, it is not really possible to access the metrics and run the PromQL query against it.
> If we could make this a minimal troubleshooting guide for the crashlooping and OOM killed case, I think that would be sufficient. WDYT?

Yeah, that's a good critique. I'll do another commit to simplify that.

djwfyi (approved):
Made my own set of changes.
This looks good to me with those in.

PR #2528, opened 2026-08-04
fix(platform generator): normalize schema required arrays from +optional

Piotr1215:
# Content Description

Normalizes the Platform reference generator's reflected `required` arrays against the Kubernetes `+optional` markers, so the schema itself is correct rather than relying on the renderer to re-derive requiredness. No generated content changes.

## Summary

The reflector infers requiredness from the absence of `,omitempty` in a field's json tag, which is not the same thing as Kubernetes requiredness. `ProjectSpec.AllowedNodeTypes` and `ProjectSpec.AllowedNodeProfiles` are the reproducible case: both document `+optional`, and neither can carry `,omitempty`, because nil means "allow every node type" while an empty list means "allow none". Serializing an empty list as absent would widen "allow none" into "allow all", which is why the source-side fix in ENGCP-969 was canceled.

This adds a normalization pass right after reflection: for every object, a property documenting a standalone `+optional` marker is dropped from its parent's `required` array.

Nothing was broken on the docs site. #2321 already taught `renderField` to demote `+optional` fields, so the published badge was correct. What was wrong was the schema underneath it, and the fact that correctness depended on the renderer re-deriving requiredness from the description. This is hardening, not a user-visible fix.

## Key Changes

`hack/platform/util/schema.go`

- `generateSchema` now runs `normalizeRequired` on the reflected schema before returning it.
- `normalizeRequired` walks the root plus every reachable subschema and strips `+optional` properties from each object's `required` array. An object left with nothing required drops `required` entirely instead of emitting an empty array.
- Nullable fields are handled: `jsonschema:"nullable"` reflects as a `oneOf` of the real type plus null, which leaves the property's own description empty and moves the marker onto the non-null member, so the check falls back to that member. This reuses the shape `renderField` already established in #2321.
- The walk is deliberately lenient. A cycle terminates via a visited set, a `required` entry naming a property that cannot be resolved is kept rather than guessed at, and an unrecognized schema shape is simply not descended into. Given how often this generator has been broken by upstream `loft-sh/api` churn, an unexpected shape should degrade to a no-op instead of failing generation.

`hack/platform/util/schema_test.go`

Four new tests. The file previously had no schema-generation coverage at all, so this adds the fixture scaffolding too. Comments are seeded directly into the package comment map, keyed the way the reflector looks them up, which keeps the tests off the vendor extraction path (it needs a `vendor` directory relative to the working directory, which does not exist under the package dir). Unknown field names fail loudly, so a rename cannot quietly turn the assertions into no-ops.

- `TestGenerateSchemaNormalizesRequiredFromOptionalMarkers`: an optional slice where nil and empty differ is demoted, and a genuinely required field stays required.
- `TestGenerateSchemaNormalizesNullableOneOfDescription`: asserts the reflector really does produce the nullable `oneOf` shape before asserting the demotion, so the test cannot pass for the wrong reason.
- `TestNormalizeRequiredClearsFullyOptionalObjects`: a fully optional object emits no `required` key.
- `TestNormalizeRequiredToleratesUnusualSchemas`: nil schema, `required` without properties, `required` naming an absent property, and a self-referential schema.

### On the renderer fallback

Issue step 5 asked whether to keep or simplify the `renderField` demotion. It stays, and it is not redundant. The vcluster generator enters this package through `RenderFromPath` with a schema loaded from `values.schema.json`, which never passes through `generateSchema` and so is never normalized. Removing the fallback would make correctness depend on the normalization pass reaching every render path, which it does not. The marker-stripping loop in `renderField` also has to stay regardless, since it removes `+` lines from the rendered prose.

## Verification

Acceptance criteria, all checked against the artifact rather than reasoned about:

- `AllowedNodeTypes` and `AllowedNodeProfiles` absent from the reflected `ProjectSpec.required`. Confirmed by reflecting the real `storagev1.ProjectSpec` at the current `go.mod` pin (v4.11.0). Before: `required = [allowedNodeTypes allowedNodeProfiles]`. After: `required = []`.
- Source json tags unchanged. No `,omitempty` added anywhere; this PR touches two files, both under `hack/platform/util`.
- Regression test for an optional slice where nil and `[]` differ. Covered.
- Test for the nullable `oneOf` description shape. Covered.
- Genuinely required fields still required. Covered.
- Generated output still marks both fields optional. Yes, and unchanged.
- No unrelated requiredness changes. Confirmed the strong way: a baseline regeneration on `main` produces zero content changes, proving the checked-in partials already match the current generator, and a regeneration with this change also produces zero content changes across all 158 files. So no field anywhere changed requiredness in the rendered output, which is the expected result given #2321.

The three behavioral tests were verified to fail with the normalization call disabled and pass with it, so they are not vacuous.

One caveat on that, worth a reviewer's attention: no workflow in this repo runs `go test` on pull requests, so these tests were run locally (`go test ./hack/platform/util/`, plus `go build ./hack/...`, `go vet`, and `gofmt`) and CI will not re-run them here or guard against a future regression. Wiring up a Go test job is separate scope and not attempted in this PR, but it does limit what the green checks below actually prove about `hack/`.

Note on file modes: regenerating rewrites 158 partials from mode 755 to 644, because the checked-in copies predate the `writeGeneratedFile` permission fix. That drift is pre-existing and unrelated to this issue, so it is deliberately not included here. Worth a separate cleanup.

## Preview Link

This PR changes no `.mdx` files, so there are no changed documents to preview. The page whose badges the change concerns, expected to be identical to production:

- https://deploy-preview-2528--vcluster-docs-site.netlify.app/docs/platform/next/api/resources/project/project

## Backporting

Not applicable. The generator source is single-copy on `main`, and the rendered output does not change, so there is nothing to apply to `platform_versioned_docs/`.

## Dependencies

None. Self-contained in one package.

## TODO

- [x] Normalization pass
- [x] Regression tests, verified non-vacuous
- [x] Regeneration diff measured against a baseline
- [ ] Review

## Internal Reference

Closes DOC-1659

@netlify /docs

### DIFF
--- a/hack/platform/util/schema.go
+++ b/hack/platform/util/schema.go
@@ -81,9 +81,10 @@ func generateSchema(configInstance interface{}) *jsonschema.Schema {
 	// than from explicit `jsonschema:"required"` tags, which those structs do
 	// not carry. renderField then derives the "required" badge from that array,
 	// so a field is only flagged required when the upstream type genuinely
-	// requires it within its parent object. The `+optional` comment override in
-	// renderField further demotes fields that lack `,omitempty` but are
-	// annotated optional.
+	// requires it within its parent object. normalizeRequired below reconciles
+	// the array with the `+optional` markers the fields document, and the same
+	// override in renderField keeps the badge correct for schemas that never go
+	// through reflection.
 	r.RequiredFromJSONSchemaTags = false
 	r.ExpandedStruct = true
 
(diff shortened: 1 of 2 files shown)


djwfyi (approved):
LGTM

PR #2641, opened 2026-08-26
docs(customresources): add a calico walkthrough for scoping synced selector fields

r0bnet:
# Content Description

Documents how to scope a synced selector field to the tenant that wrote it, using Calico `NetworkPolicy` as the worked example. A selector held as an opaque string can't be translated, so vCluster copies it verbatim and it then matches beyond the tenant namespace it was written for. Adds a new tutorial page plus a short note and links from the to-host config reference and the manage page.

The `vcluster.yaml` snippet is byte-identical to the fixture covered by the `e2e/calico` lane in loft-sh/vcluster-pro#2310, so the config on the page is regression-tested rather than illustrative.

## Preview Link

- https://deploy-preview-2641--vcluster-docs-site.netlify.app/docs/vcluster/next/learn-how-to/sync-calico-network-policies

## Internal Reference

ENGCP-1387 (this documents a workaround, the underlying issue stays open)

@netlify /docs

### DIFF
--- a/vcluster/configure/vcluster-yaml/sync/patching/README.mdx
+++ b/vcluster/configure/vcluster-yaml/sync/patching/README.mdx
@@ -52,6 +52,10 @@ patches:
     expression: 'value.replace("internal.svc", "internal.svc.cluster.local")'
 ```
 
+:::warning Omitting expression on an array-nested path
+Omitting `expression` or `reverseExpression` drops that value from the patch, which normally leaves the target's existing value alone. Arrays are the exception. JSON Merge Patch replaces an array wholesale whenever it appears in the patch. If the array shows up in the patch for any other reason, dropping one nested value doesn't preserve the rest of the target's array. It overwrites the whole array without them. Remove the array's own key from the patch instead of a value nested inside it, if you need the target's array left untouched.
+:::
+
 ## Patch modes
 
 Choose the mode based on what the field contains:
@@ -148,7 +152,7 @@ Expressions can use these helper functions:
 
 ### Empty paths
 
-Empty paths are an advanced use case. Use them when you need to read multiple fields to decide what to change, or when you need to set multiple fields in a single expression.
+Empty paths are an advanced use case. Use them when you need to read multiple fields to decide what to change, or when you need to set multiple fields in a single expression. They're also the only way to set a field a path-specific patch can't reach, because that field isn't present in the change being synced.
 
 With a normal path, `value` is the single field the patch targets. With an empty path, `value` is the entire set of fields being changed in this sync operation. This is only the changed fields, not the full object. The expression must return the modified change set.
 
--- a/vcluster/configure/vcluster-yaml/sync/to-host/advanced/custom-resources.mdx
+++ b/vcluster/configure/vcluster-yaml/sync/to-host/advanced/custom-resources.mdx
@@ -127,6 +127,8 @@ sync:
           reverseExpression: "value.startsWith('prod-') ? value.slice(5) : value"
 ```
 
+A selector held as an opaque string can't be translated, so vCluster copies it verbatim. Every tenant namespace maps to a single namespace on the control plane cluster, so the copied selector matches objects the policy was never written for. An expression patch can scope the field to the namespace it came from. For a worked example, see [Scope a synced selector field to its tenant](../../../../../learn-how-to/sync-calico-network-policies.mdx).
+
 ## Sync an Istio waypoint Gateway using custom resources
 
 :::tip Use native Gateway API sync for core Gateway API resources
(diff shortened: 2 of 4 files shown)


djwfyi:
Ran a full accuracy + readability pass on this guide (source-verified against `vcluster-pro`, several claims confirmed with throwaway Go tests against the actual patch engine). Summary of what changed on top of the original commit:

## Fixes

- **Reverse-sync leak (the important one).** The empty-path patch composed a scoped `spec.selector` on the way to the control plane cluster, but nothing stripped it back out on the way back to the tenant. Traced the actual code path this resource uses (`genericsyncer.go`'s `virtualPatch = CalculateMergePatch(event.HostOld, event.Host)`) and reproduced the leak with a test: once the composed selector lands in the host object, a later reconcile (cold cache, restart, anything that makes the syncer re-diff the host object) can read it back as a "host change" and write it straight into the tenant object, undoing the whole point of the guide.
  - First fix attempt (restore a plain `spec.selector` patch with no `reverseExpression`) turned out to be worse — deleting a nested key inside the `spec.ingress` array doesn't stop JSON Merge Patch's atomic array replacement, so it was erasing the tenant's own rule selectors instead of leaving them alone. Verified that with a test too.
  - Real fix: the empty-path patch's `reverseExpression` now deletes `spec.selector`, `spec.ingress`, and `spec.egress` as whole fields (not nested values), so neither bug applies. Confirmed with a test that forward composition, reverse cleanup, and an unrelated host-side annotation change all behave correctly together.

- **Wrong FeatureTable key.** Was `vcp-distro-sync-patches`. The actual code path for `customResources` patches (`pkg/customresources/to_host_syncer.go`) bypasses the license check that key maps to entirely — it's gated by `VirtualClusterProDistroGenericSync` instead, checked unconditionally at syncer registration. Fixed to `vcp-distro-generic-sync`, and added the missing `pro` sidebar tag + `ProAdmonition` (this is a Pro Distro feature and the guide wasn't disclosing that).

- **RBAC narrowed to what's actually used.** Added `resourceNames` scoping (`tier.networkpolicies/default.*`, `tiers/default`) matching Calico's own documented tiered-policy RBAC model, and dropped `tier.globalnetworkpolicies`/`list`/`watch` on `tiers` since this guide never syncs GlobalNetworkPolicy and only ever needs `get` on the tier.

- **"One tenant cluster per namespace" claim corrected.** The original wording was ambiguous about whether multi-tenant-per-namespace was ever a safely-supported configuration. It wasn't — it was an experimental `reuseNamespace` flag, deprecated on introduction and removed in v0.25 (PR #2713). Reworded, and moved it to sit next to the related Prerequisites bullet about namespace-sync mode instead of floating disconnected from it earlier in the page.

## Coverage gaps closed

- Extended patches from selector-only to all five locations Calico actually reads: top-level `spec.selector` plus the four rule-level `source`/`destination` selectors on ingress/egress. The original only patched the top-level selector.
- New Limitations bullet: a rule with no `source`/`destination` selector at all isn't scoped either, and can't be — Calico matches an empty entity rule against every endpoint in the policy's own (flattened, multi-tenant) namespace, the same leak this guide exists to prevent, and a path-specific patch can't add a field a rule doesn't have.
- Walkthrough now proves the positive case too (client in `demo` can still reach the workload it's allowed to reach), not just the negative case, and the final round-trip check now reads back the rule-level selector as well as the top-level one — directly exercising the reverse-leak fix above instead of only checking half of it.

## Style / readability

- Converted the `HOST_CTX`/`VCLUSTER_CTX` export block to `InterpolatedCodeBlock` (editable in the rendered docs); filed [DOC-1721](https://linear.app/loft/issue/DOC-1721) to do the same for the other guides using this pattern.
- Added YAML inline comments to the patch list (missing entirely before, inconsistent with `patching/README.mdx` and the sibling guide).
- Dropped an `exit 1` from the cross-namespace check — it would kill the reader's shell if pasted into an interactive terminal and the check raced Calico's async policy programming, which the guide's own next sentence warns is possible.
- Split an over-length sentence and trimmed a YAML comment that had grown to duplicate the prose below it.

## Follow-ups filed (not blocking, separate from this guide)

- [DOC-1722](https://linear.app/loft/issue/DOC-1722) — `patching/README.mdx`'s own empty-path example uses `??=`, which the vendored goja build doesn't support (confirmed with a test — throws `SyntaxError`).
- [DOC-1723](https://linear.app/loft/issue/DOC-1723) — a resource's `patches` apply in reverse configuration-list order (verified in `pkg/translate/translate.go`), which isn't documented anywhere in the patching reference.

Vale, config-drift, and feature-table validation all pass.

djwfyi (approved):
I've my changes look good to you, then this looks good to me.

PR #2785, opened 2026-09-14
fix(config-partials): resolve stale lowercase resourceClaims/resourceClaimTemplates partials

djwfyi:
# Content Description
Versioned DRA docs (0.35.0, 0.36.0, 0.37.0) imported lowercase `resourceclaims.mdx`/`resourceclaimtemplates.mdx` partials that the config-partials generator no longer writes to since PR #2776 (DOC-1739) fixed its `paths` allowlist to use camelCase filenames, matching every other generated partial and `main`'s already-correct imports. That left the versioned pages serving permanently stale content while the generator's real output landed in dead, unimported camelCase files — grandfathered into `legacyOrphanFiles` to keep the build green rather than fixed. This repoints the six versioned DRA imports at the camelCase partials, deletes the six now-dead lowercase files, and drops the two obsolete `legacyOrphanFiles` entries.

## Preview Link
Netlify will post the deploy preview on this PR.

## Internal Reference
Closes DOC-1776


djwfyi:
Confirmed and fixed in 25c80ae. Only 0.37.0 had camelCase partials generated — I'd only checked that version before deleting the lowercase files, missing that 0.35.0/0.36.0 never had them.

Since the pre-deletion lowercase content was byte-identical across all three versions (schema unchanged since 0.35.0), I added the already-regenerated 0.37.0 camelCase content to 0.35.0/0.36.0 rather than reintroducing the stale pre-#2776 content (which had wrong `required` flags on several fields).

Also addressed the non-blocking README/comment note — only `sleepMode.mdx` remains in the legacy-orphan exemption now.

djwfyi:
Verified this properly rather than resting on the byte-identical-stale-files inference: pulled the real \`chart/values.schema.json\`/\`values.yaml\` from loft-sh/vcluster at tags \`v0.35.0\`, \`v0.36.0\`, \`v0.37.0\`, ran the actual generator against each, and diffed the output against what's committed here.

Zero differences, all three versions, both files. The 0.35.0/0.36.0 content added in 25c80ae matches what the generator produces from each version's real upstream schema — the DRA resourceClaims/resourceClaimTemplates config surface hasn't changed since 0.35.0, confirmed against source rather than assumed from the stale copies agreeing with each other.

PR #2788, opened 2026-09-14
fix(certified-stacks): document the real customization flow

djwfyi:
# Content Description

The "Customize a certified stack" sections told users to copy or duplicate a certified StackTemplate, but the UI has no copy or duplicate action. Replaced that guidance with the flow that works today, then restructured the section so each interface stands on its own.

**The fix.** Certified templates expose only a **View** action, and **Save Changes** is permanently disabled. The docs now describe copying the YAML from the **CR** tab and creating a new template from it, including the metadata that has to change.

**Two self-contained tabs.** The management API path previously trailed off the end of the UI procedure and said to edit the file "the same way," so anyone working outside the UI had to read UI steps that didn't apply. Platform UI and Management API are now separate tabs, each complete on its own.

**Built for a growing catalog.** Run:ai is the only certified bundle today, so the old text hardcoded its template name and asserted facts about that one bundle. The kubectl path now lists templates before exporting one, identifies certified templates by the **Certified** badge or the `vcluster.com/certified` annotation, and uses an `InterpolatedCodeBlock` for the template name. The integrity guidance now explains the mechanism (task names appear in `dependsOn` edges and `{{ .Outputs.<task>.<output> }}` references, outputs can only read namespaces the Stack deployed into, chart versions are pinned to a tested combination) rather than stating that the Run:ai bundle relies on those four things.

Verified against `loft-enterprise`:

- `ui/src/views/Templates/StackTemplates/StackActionsCell.tsx` returns only a **View** action for a certified template. No Edit, Delete, Show YAML, or Duplicate.
- `EditStackTemplatePage` opens that template with `CERTIFIED_READ_ONLY_REASON`, which disables **Save Changes**.
- Both the View and Create pages render `YAMLFocusedForm` with the editor pane open, so the **CR** tab is where the YAML gets copied and pasted.
- `StackNameCell.tsx` renders the **Certified** badge beside the template name in the Stacks list.
- `pkg/defaults/stacktemplates/certified/certified-stacks.yaml` sets `vcluster.com/certified` on all four Run:ai StackTemplates. There is no label, so `-l` selectors can't filter for them.
- `pkg/defaults/defaults.go` re-applies bundled objects when the certified bundle generation increases, which is why a separately named template is the safe place for changes.

Engineering plans to add a Duplicate action, which would land in a future stable release. This PR documents current behavior only; the UI tab needs a revisit when that ships.

## Preview Link

- https://deploy-preview-2788--vcluster-docs.netlify.app/docs/platform/next/administer/templates/certified-stacks#customize-a-certified-stack
- https://deploy-preview-2788--vcluster-docs.netlify.app/docs/platform/next/integrations/certified-stacks/#customize-a-certified-stack

## Internal Reference

Closes DOC-1777

@netlify /docs

🤖 Generated with [Claude Code](https://claude.com/claude-code)


cbron (approved):
Fine as temporary fix.

